PR #4, opened 2019-12-11
Scaffold Week 6 (Authentication) Learning Outcomes

oliverjam:
I skipped Week 5 for now as DBs aren't my favourite. Hopefully we can establish this structure so each week can mirror it.

```markdown
# Week X Learning Outcomes

## Specific week LOs

## Design LOs

## Testing LOs

## Developer Toolkit LOs
```

### DIFF
--- a/week-6/learning-outcomes.md
+++ b/week-6/learning-outcomes.md
@@ -4,12 +4,13 @@
 
 ### Password encryption
 
+- [ ] **We can store passwords securely**
 - [ ] We can hash and salt passwords using bcrypt
 - [ ] We describe why it's important to hash and salt passwords
 
 ### Authentication
 
-- [ ] We can authenticate users by setting and reading cookies
+- [ ] **We can authenticate users by setting and reading cookies**
 - [ ] We can use JSON Web Tokens to store user information in a cookie
 - [ ] We can describe why we need to use JSON Web Tokens
 - [ ] We can ensure only authenticated users can see certain content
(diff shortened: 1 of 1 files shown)


week-6/learning-outcomes.md
```diff
@@ -4,12 +4,13 @@
 
 ### Password encryption
 
+- [ ] **We can store passwords securely**
 - [ ] We can hash and salt passwords using bcrypt
 - [ ] We describe why it's important to hash and salt passwords
```
sofer: Not sure how to test for this. How about, *"We can demonstrate how to execute a successful attack on an unencrypted password field"*? (or is that a stretch goal?)

week-6/learning-outcomes.md
```diff
@@ -18,36 +19,36 @@
...
 
-- [ ] We can measure the test coverage of our code
+- [ ] **We can measure the test coverage of our code**
 - [ ] We can identify where our tests are lacking and fix this
 
 ### Static analysis
 
-- [ ] We can configure ESLint to analyse our code
+- [ ] **We can configure ESLint to analyse our code**
 - [ ] We can describe how a linter helps us write code
```
sofer: How would we test for this? Is something like *"We can demonstrate a failed git merge due to inconsistencies in formatting"* too silly?

oliverjam: The idea of this being in the "Testing" section was showing how ESLint can prevent actual runtime bugs (e.g. by catching an undefined variable). It's as close as you can get to typechecking vanilla JS 😆

I would say Prettier is a much better choice for _formatting_ at this point.

If we wanted a more specific outcome we could use "We can configure ESLint to catch common JS errors"?

week-6/learning-outcomes.md
oliverjam: Hmm that might be a bit much for them at this point. That would probably come under "describe why hashing is important".

The "top-level" bolded goals are all slightly more generic as they're meant to represent the overarching theme for that section. This perhaps makes them a bit less testable, so maybe they should instead be used as headings or something?

PR #5, opened 2019-12-11
Scaffold Week 7 (REST APIs) Learning Outcomes

oliverjam:


week-7/learning-outcomes.md
```diff
@@ -0,0 +1,47 @@
...
+
+- [ ] **We can structure our server code effectively**
+- [ ] We can structure our code to match REST methods
+
+### Express
+
+- [ ] **We can use Express to build an HTTP server**
+- [ ] We can serve our static assets using Express
+- [ ] We can use Express to handle requests for different routes
+- [ ] We can explain what Express middleware is
```
sofer: How would we review this learning outcome?

week-7/learning-outcomes.md
```diff
@@ -0,0 +1,47 @@
...
+- [ ] We can use 3rd party middleware
```
sofer: Is it worth being more specific?, e.g. *"We can use at least three pieces of third party middleware in our app"*.

oliverjam: I wouldn't want to give them the idea that there's a minimum amount of 3rd party stuff they should be using. I think I originally had more specific outcomes like "can use middleware to handle errors", "can use middleware to parse POST bodies" etc. That might be a good option

week-7/learning-outcomes.md
oliverjam: All of the "we can explain/describe" ones are tough to review. I see those as targeted more at the student themselves (so they know what do and don't understand), but maybe we can find a way.

oliverjam: I think this is covered by the more specific LOs right below: if you can use `body-parser` and write your own error-handling middleware then I think you've demonstrated that you understand middleware

PR #6, opened 2019-12-11
Scaffold Week 8 (Full stack app) Learning Outcomes

oliverjam:


week-8/learning-outcomes.md
```diff
@@ -0,0 +1,28 @@
...
+- [ ] We can render elements in the DOM based on the server data
+
+### Error handling
+
+- [ ] We can respond with useful error information to the client
+- [ ] We can throw errors when our server hits a problem
+- [ ] We can catch errors and log them correctly
+- [ ] We can handle errors that occur in async code
+- [ ] We can show the user useful messages based on server errors
+
```
sofer: Is it asking too much that, *"We can write automated tests for our error handling"*?

oliverjam: No I think that's a great idea. They should know how to mock by this point, so it wouldn't be unreasonable to ask them to mock an error response from their API and test that they show a relevant message to the user.

PR #7, opened 2019-12-12
Scaffold Week 5 (Databases) Learning Outcomes

oliverjam:
Still working on the rest

### DIFF
--- a/week-5/learning-outcomes.md
+++ b/week-5/learning-outcomes.md
@@ -1,6 +1,6 @@
 # Week 5 Learning Outcomes
 
-## Database design
+## Databases
 
 - [ ] We can design and document a database schema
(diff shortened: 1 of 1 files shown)


bobbysebolao:
Looking good so far! I think this would also be a good week to introduce the software architecture concepts that are currently covered in week 3, students will be drawing db schema and building a larger app than in any of the previous weeks. A possible learning outcome could be:

**We can document the architecture of our app**

- [ ] We can draw a diagram to represent the overall architecture
- [ ] We can create and document a schema for a relational database
- [ ] We can follow a naming convention for CSS classes, ids, JS variables, files, and folders

oliverjam:
@bobbysebolao I've added a couple of architecture LOs. The DB schema one was already covered in the Databases section

PR #8, opened 2019-12-13
Add week 4 learning outcomes

bobbysebolao:


learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] We can ‘chunk’ data sent from the client side on the server side
+- [ ] We can give descriptive names to HTML form input fields so that it is easy to access data in the POST request object
+- [ ] We can parse a request object
+- [ ] We can stringify a response object
+
+### Core modules
+
+- [ ] **We can use core Node modules**
+- [ ] We can require core modules in our files
+- [ ] We can access and invoke methods provided by core modules
```
oliverjam: These three kind of say the same thing

learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] We can read files synchronously and asynchronously using Node’s fs module
+
+## Design LOs
+
+- [ ] **We can design a website that is easy to use and to navigate**
+- [ ] We can identify common design patterns that aid usability
+- [ ] We can build website navigation that demonstrates good usability
+- [ ] We can make HTML elements look different on hover and focus to indicate their state to the user
+
+## Testing LOs
```
```suggestion
## Testing
```

learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] We can stringify a response object
+
+### Core modules
+
+- [ ] **We can use core Node modules**
+- [ ] We can require core modules in our files
+- [ ] We can access and invoke methods provided by core modules
+- [ ] We can read files synchronously and asynchronously using Node’s fs module
+
+## Design LOs
```
```suggestion
## Design
```

learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] We can serve different types of files to the client
+- [ ] We can read information sent in a URL query string
+
+### POST requests
+
+- [ ] We can handle data received from a POST request
+- [ ] We can ‘chunk’ data sent from the client side on the server side
+- [ ] We can give descriptive names to HTML form input fields so that it is easy to access data in the POST request object
+- [ ] We can parse a request object
+- [ ] We can stringify a response object
```
oliverjam: What exactly does this mean? Responding to the browser with an HTML/JSON string?

Req/res stuff isn't really specific to POST requests, so maybe we don't want them in this section?

learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] We can read secrets from a .env file
+- [ ] We can prevent sensitive data from being pushed to GitHub
+
+### GET requests
+
+- [ ] We can route requests to the correct handler function(s)
+- [ ] We can serve different types of files to the client
+- [ ] We can read information sent in a URL query string
+
+### POST requests
```
oliverjam: The `GET` section is stuff that applies to any HTTP method really. Most of the `POST` stuff is about handling HTTP bodies, which lots of methods can have.

Could we instead put the `GET` stuff under "Servers" and have the specific "submitting data" LOs under "Request bodies" or something?

learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+
+- [ ] We can handle data received from a POST request
+- [ ] We can ‘chunk’ data sent from the client side on the server side
```
oliverjam: This might be misleading—the request body arrives in chunks that we have to handle. We don't actually chunk the data ourselves

```suggestion
- [ ] We can handle streams of data sent from the client that don't arrive all at once
```

learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] We can give descriptive names to HTML form input fields so that it is easy to access data in the POST request object
```
```suggestion
- [ ] We can give descriptive names to HTML form input fields so that it is easy to access data in the request
```
Forms make `GET` requests by default (see [my new workshop](https://github.com/oliverjam/learn-html-forms)) so this should probably be a more general LO

week-4/learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+### Servers
+
+- [ ] **We can create a web server that responds to http requests**
+- [ ] We can conditionally set ports for our server based on the runtime environment
+- [ ] We can read secrets from a .env file
+- [ ] We can prevent sensitive data from being pushed to GitHub
+
+### GET requests
+
+- [ ] We can route requests to the correct handler function(s)
```
arrested-developer: This isn't specific to GET requests, should this be moved up to the general "Servers" section?

week-4/learning-outcomes.md
```diff
@@ -0,0 +1,50 @@
...
+- [ ] We can serve different types of files to the client
+- [ ] We can read information sent in a URL query string
+
+### POST requests
+
+- [ ] We can handle data received from a POST request
+- [ ] We can ‘chunk’ data sent from the client side on the server side
+- [ ] We can give descriptive names to HTML form input fields so that it is easy to access data in the POST request object
+- [ ] We can parse a request object
```
arrested-developer: Does this mean parse the request body? I'm guessing you don't want to parse the whole request object?

arrested-developer: Maybe

- We can parse JSON sent in an HTTP request
- We can stringify JSON and send it via the HTTP response

oliverjam: I don't think we should specify JSON, since forms submit urlencoded data by default (or if we do we should specify both and make it clear when they would use each)

week-4/learning-outcomes.md
bobbysebolao: Oli pointed this out too, I've moved it into the 'Servers' section 👍

week-4/learning-outcomes.md
bobbysebolao: @oliverjam Ok, in that case, how about the following:

- We can parse JSON-encoded data sent in a **fetch request**
- We can stringify JSON and send it via the HTTP response

I think specifying that it's a fetch request avoids the confusion caused by my previous phrasing. Fetch requests send JSON-encoded data.

Surely we can get students to make both types of request (JSON-encoded fetch and default urlencoded form submission) in this week's project to build a blogging platform? i.e. when you publish a new post it makes sense for a full page reload to occur, so you would use the default form submission, but when you comment on an existing post (stretch goal), it's nice when just the DOM updates without a full page reload (so you'd use fetch).

oliverjam: Yeah that's not a bad idea. I just worry that it might get confusing for people—I think there's already a bit of difficulty understanding the difference between browser and server. I was going to suggest that we say "no client-side JS at all" for this project. We can leave the client-side requests for later weeks

oliverjam: They'll still have to do the whole body-parsing streams song and dance, it's just the last step that changes from `JSON.parse(data)` to `querystring.parse(data)` 🤷‍♂. Probably not enough of a difference to worry about

bobbysebolao: nice, that sounds like a great way to do it to me! Thanks @arrested-developer @oliverjam for the comments. So, this learning outcome will just be:

- We can parse data sent via a default HTML form submission

And we can make it clear in the project description that they shouldn't be using JavaScript to send the data for this week's project. Then in a later week we can introduce sending/parsing fetch request data.

I'll make another commit to update this 👍

PR #9, opened 2019-12-13
Add week 3 learning outcomes

bobbysebolao:


learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+- [ ] **We can test the functionality of our app**
+- [ ] We can write tests to mimic the behaviour of a user performing different actions
+- [ ] We can check that passing a given input into our tests returns the expected output
+
+### Modular functions
+
+- [ ] **We can write testable, modular functions**
+- [ ] We can write functions that add DOM nodes
+- [ ] We can write functions that remove DOM nodes
+- [ ] We can write functions that modify DOM nodes
```
oliverjam: Why the specific focus on DOM nodes in the modular functions section?

learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+
+### Event listeners
+
+- [ ] **We can apply click event listeners to buttons**
```
oliverjam: This might be too specific? They should probably be using a form `"submit"` event for the project this week for example

learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+- [ ] We can pass in named functions as the second argument
+
+### Scope
+
+- [ ] **We can apply the concept of ‘scoping’ to functions and variables**
```
oliverjam: "apply the concept" might be a little vague/confusing. Can we think of a concrete application for scoping? Like "We can use function scope to..."

learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+- [ ] We can distinguish between block, function, and global scope so that we can control access to variables from different parts of our project
+
+---
+
+## Design
+
+- [ ] **We can create a clean colour palette for our project**
```
oliverjam: I'm not sure "clean" is a very measurable outcome 😆 
How about
```suggestion
- [ ] **We can create an attractive and accessible colour palette for our project**
```

learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+- [ ] We can apply colours to HTML elements using CSS variables
+
+---
+
+## Developer Toolkit
+
+- [ ] **We can run automated tests in the browser (to check that our client side code works correctly)**
```
oliverjam: This isn't really a dev tooling outcome (other than that the tests will use `console.log`)

learning-outcomes.md
bobbysebolao: I was thinking of the to-do list project. Perhaps these would be more suited to learning outcomes for the project though, rather than the week.

learning-outcomes.md
bobbysebolao: How about "We can use function and block scope to store values inside temporary variables"?

I agree, the more concrete examples we can scaffold these LOs around, the better

week-3/learning-outcomes.md
```diff
@@ -0,0 +1,26 @@
...
+- [ ] We can write tests to mimic the behaviour of a user performing different actions
+- [ ] We can check that passing a given input into our tests returns the expected output
+
+### Modular functions
+
+- [ ] **We can write testable, modular functions**
+
+### HTML forms
+
+- [ ] **We can submit an HTML form without using JavaScript**
```
oliverjam: This is probably a better LO for week 1. This week they aren't going to have a server and will be making `fetch` requests in JS. My previous comment was just about using a "click" event rather than a "submit" event

```diff
- button.addEventListener("click", handle)
+ form.addEventListener("submit", handle)`
```

week-3/learning-outcomes.md
```diff
@@ -0,0 +1,26 @@
...
+
+### Scope
+
+- [ ] **We can use function and block scope to store values inside temporary variables**
```
oliverjam: I think this might be confusing for them. I don't know if "temporary" will make sense for any variables at this point.

Scope is mostly about isolation, so maybe something like:

```suggestion
- [ ] **We can isolate variables inside function and block scopes**
```

learning-outcomes.md
oliverjam: I don't think those are bad outcomes for this week, I just don't know if they fit under "modular functions"

bobbysebolao: I have changed the heading from 'Modular functions' to just 'Functions' to encompass all of these LOs

week-3/learning-outcomes.md
bobbysebolao: ok, that makes sense. I'm thinking something like "We can apply event listeners to HTML form elements" then.

learning-outcomes.md
oliverjam: We have a general "Javascript" heading in some weeks. Could we use that here and move the scope LO under it too?

PR #10, opened 2019-12-13
Add week 2 learning outcomes

bobbysebolao:


learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+## HTTP
+
+### Asynchronicity
+
+- [ ] **We can write code that executes asynchronously**
+- [ ] We can use callbacks to make our code non blocking
+
+### Fetch
+
+- [ ] **We can use the fetch method to make API requests and receive responses**
```
```suggestion
- [ ] **We can use the fetch method to make HTTP requests and receive responses**
```
"API" is used for so much stuff it gets confusing. Also you can make requests to servers that wouldn't necessarily be what most think of as an API

learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
+# Week 2 Learning Outcomes
+
+## HTTP
+
+### Asynchronicity
+
+- [ ] **We can write code that executes asynchronously**
+- [ ] We can use callbacks to make our code non blocking
```
oliverjam: The callback itself doesn't make code non-blocking, it just allows us to access a value later on from code that got pushed to the back of the queue. Also we should have promises here if they're going to be using fetch

```suggestion
- [ ] We can use callbacks to access values that aren't available synchronously
- [ ] We can use promises to access values that aren't available synchronously
```
```

File: learning-outcomes.md [hunk 3]
```diff
@@ -0,0 +1,38 @@
...
+- [ ] We can use an API key to bypass rate limit restrictions on certain APIs
+
+- [ ] **We can use the map and filter methods to display different data based on user input**
+- [ ] We can return filtered items from an array that match specified criteria
+- [ ] We can map over an array to create a new array containing new values
+
+- [ ] **We can manipulate the DOM using JavaScript**
+- [ ] We can use different selector methods to select DOM nodes
+- [ ] We can add and remove DOM nodes to change the appearance of our website
+- [ ] We can toggle the classes applied to DOM nodes to change their properties
```
oliverjam: Does "properties" here refer to CSS properties?

learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Design
+
+- [ ] **We can use consistent layout and spacing**
+- [ ] We can implement readability margins to ensure a comfortable reading experience on different devices
```
oliverjam: What does "readability margins" mean? Preventing the line-length from getting too long or stopping the text hitting the edge of the page? (Or both?)

learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+- [ ] We can follow a spacing guideline to give our app a consistent feel
+- [ ] We can strike a good balance between active space and white space in our app
+
+## Developer Toolkit LOs
+
+- [ ] **We can debug client side JS in our web browser**
+- [ ] We can inspect the DOM in our browser to see which elements are there
+- [ ] We can view the console in our browser to help us debug our JavaScript
+- [ ] We can use console.log() to help us debug our code
```
oliverjam: I think these two are basically the same?

learning-outcomes.md
```diff
@@ -0,0 +1,38 @@
...
+- [ ] We can setup Prettier so that our code is more readable
```
oliverjam: Bit pedantic sorry, but "setup" is a noun. "set up" is the verb. Also I'd say the most important Prettier benefit to highlight is consistency within a team

```suggestion
- [ ] We can set up Prettier so that our formatting is consistent across the team
```

learning-outcomes.md
bobbysebolao: Yeah, I'll make that explicitly clear though

learning-outcomes.md
bobbysebolao: I meant preventing the line length from getting too long. I'll rephrase to make it clearer

learning-outcomes.md
bobbysebolao: That's a great point, I agree that's the most important benefit of using Prettier

week-2/learning-outcomes.md
```diff
@@ -0,0 +1,37 @@
+# Week 2 Learning Outcomes
+
+## HTTP
+
+### Asynchronicity
+
+- [ ] **We can write code that executes asynchronously**
+- [ ] We can use callbacks to access values that aren't available synchronously
+- [ ] We can use promises to access values that aren't available synchronously
```
arrested-developer: We need to give students a while to understand callbacks before we move onto promises IMO

arrested-developer: Having tried both ways, I think that our standard approach of starting with XMLHttpRequest is the better option, as it makes it clearer what is happening

oliverjam: Do you mean you tried starting with `fetch` at Beamery? The middle east campuses have been doing first-class functions -> callbacks -> async -> fetch/promises all in one day for a long time now and it seems to work for them.

What concepts does XMLHttpRequest make clearer?

arrested-developer: Yeah, I tried starting with fetch, but felt like it abstracted away too much of what I wanted to get across. With XHR I like how the callback is called on each readystate change and how this demonstrates the lifecycle of the HTTP request. But none of that is exactly necessary...

>The middle east campuses have been doing first-class functions -> callbacks -> async -> fetch/promises all in one day for a long time now and it seems to work for them.

I think the advantage to this was to take a simpler option than using XHRs in the early stage of the course. We still need to map out how we build up understanding and concepts around async. By using promises so early, I worry that we don't give students time to understand or use callbacks.

oliverjam: Yeah I understand your concern. At this point though the JS ecosystem is so saturated in promises that we lose more by not covering them early on. I don't think it's necessarily a bad thing to get higher-level concepts early on and understand more deeply once you're comfortable with getting stuff done.

Also promises still use callbacks, so the basic concept has to be understood.

What exactly were you teaching around the lifecycle of the HTTP request? We might still be able to cover how it works (or what fetch is doing under-the-hood)

bobbysebolao: @arrested-developer It's a good point because IMO callbacks are hard to grasp at first, but important to understand (even if using promises is the norm for many tasks nowadays).

@oliverjam What about if we put a workshop in this week that shows students how to use both callbacks and promises to complete the same task? Like how the error handling workshop in week 5 gets you to write the same set of functions, but with different error handling methods?

I think it would also be worth making callbacks its own research topic, there was a really good FAC18 presentation that explained the node error first callback pattern to the group.

bobbysebolao:
@oliverjam @arrested-developer @sofer I think, once there is agreement about how to teach callbacks and promises, this will be ready to merge. 

I suggest we teach both in week 2, similar to how [this workshop](https://github.com/foundersandcoders/error-handling-workshop/blob/master/docs/approach_1.md) in week 5 teaches you 3 error handling methods (and then it's up to you which one you use in your end of week project)

week-2/learning-outcomes.md
oliverjam: I think callbacks are too important and too confusing (initially) to leave to a research topic. It's hard enough to explain even when you know exactly how to use them yourself.

I honestly don't think it will be that big an issue to have promises and callbacks side by side. The most important thing is for them to consistently get exposed to and practice using them throughout the course, so it clicks by the end.

sofer: @arrested-developer I understand why you want to keep xhr for conceptual understanding. On the other hand, it might be a good idea to expose a cohort to promises before the testing and server weeks (possibly the the most stressful weeks of the programme). Also, going with fetch will keep the Middle East campuses on board, which I am keen to do.

PR #11, opened 2019-12-13
Add week 1 learning outcomes

bobbysebolao:


learning-outcomes.md
```diff
@@ -0,0 +1,58 @@
...
+
+### Accessibility
+
+- [ ] **We can make a website that is accessible by people other than ourselves**
+- [ ] We can structure a site using semantic HTML to aid accessibility
+- [ ] We can add appropriate attributes to HTML elements to make them screen reader friendly
+- [ ] We can add ARIA labels (where necessary) to make our semantic HTML more accessible
+- [ ] We can run a Lighthouse accessibility audit to check our website’s accessibility
+- [ ] We can use a screen reader to understand how people with visual impairments maye experience our website
+- [ ] We can implement a suitable contrast ratio on our website so that it meets the needs of people who are colour blind
```
oliverjam: I think these are two related but separate concepts. Higher contrast benefits everyone, especially people with visual impairments. Colour-blindness is more affected by the hues chosen rather than contrast levels.

```suggestion
- [ ] We can design a UI without relying solely on colour, so that we don't exclude colour-blind users
- [ ] We can ensure our UI has sufficient colour contrast so that everyone can perceive it comfortably
```

learning-outcomes.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] We can use various tools (Lighthouse, Chromevox, WCAG Contract Checker Extension, Devtools) to check that our website meets accessibility criteria
```
oliverjam: Chromevox is a weird one. I don't think anyone actually uses it as assistive tech—it seems to only be used by devs doing testing. It's ok for getting the general idea of what a screenreader is, but they're all so different and have such specific requirements that it seems like a waste to teach something that no one really uses.

Would it be unfair to ask Mac users to use Voiceover and have Linux users stick with ChromeVox? (Since JAWS is v expensive)

```suggestion
- [ ] We can use various tools (Lighthouse, Chromevox, WCAG Contrast Checker Extension, Devtools) to check that our website meets accessibility criteria
```
Typo

learning-outcomes.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] **We can make our website display well on screens of different sizes**
+- [ ] We can use CSS media queries to change the layout on different devices/viewports
+- [ ] We can design mobile-first to ensure mobile users have a great experience on our site
+
+---
+
+### Grid layout
+
+- [ ] **We can use CSS Grid to apply responsive layouts to our website**
+- [ ] We can use an 8pt grid to create a consistent layout for our web pages
```
oliverjam: Could we emphasise that this is a design concept?

```suggestion
- [ ] We can use an 8pt grid to design a consistent layout for our web pages
```

learning-outcomes.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] We can use various tools (Lighthouse, Chromevox, WCAG Contract Checker Extension, Devtools) to check that our website meets accessibility criteria
+
+---
+
+## Design
+
+### Responsive design
+
+- [ ] **We can make our website display well on screens of different sizes**
+- [ ] We can use CSS media queries to change the layout on different devices/viewports
```
oliverjam: I think this might reinforce the idea that media queries should be specific-device-first, rather than content-first.

```suggestion
- [ ] We can use CSS media queries to ensure our content is always presented effectively
```

learning-outcomes.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] We can design mobile-first to ensure mobile users have a great experience on our site
```
oliverjam: I'm not sure how to phrase it, but imo mobile-first is just an easier way to create a responsive layout (both easier to design and code).

learning-outcomes.md
bobbysebolao: That's interesting - I'll omit to specify the screenreader that should be used then

learning-outcomes.md
bobbysebolao: Good point, agreed.

learning-outcomes.md
bobbysebolao: Perhaps "We can demonstrate a mobile-first approach to designing a website with a great user experience"

oliverjam: Yeah maybe. I wouldn't mind also clarifying that it's not "mobile-only" (this is something we used to joke about at Ticketmaster after seeing all the FAC projects).

learning-outcomes.md
```diff
@@ -0,0 +1,58 @@
...
+
+---
+
+### Accessibility
+
+- [ ] **We can make a website that is accessible by people other than ourselves**
+- [ ] We can structure a site using semantic HTML to aid accessibility
+- [ ] We can add appropriate attributes to HTML elements to make them screen reader friendly
+- [ ] We can add ARIA labels (where necessary) to make our semantic HTML more accessible
+- [ ] We can run a Lighthouse accessibility audit to check our website’s accessibility
```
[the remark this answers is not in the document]
bobbysebolao: Ok, I've changed that

learning-outcomes.md
```diff
@@ -0,0 +1,58 @@
...
+- [ ] We can use a screen reader to understand how people with visual impairments maye experience our website
```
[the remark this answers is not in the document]
bobbysebolao: Thanks, fixed

learning-outcomes.md
bobbysebolao: Have accepted your suggested change.

learning-outcomes.md
bobbysebolao: Fixed 👍

learning-outcomes.md
bobbysebolao: Yep, this LO is written in the Design subsection, I've added an extra clarification which I think makes it clear

learning-outcomes.md
```diff
@@ -0,0 +1,59 @@
...
+
+---
+
+### Accessibility
+
+- [ ] **We can make a website that is accessible by people other than ourselves**
+- [ ] We can structure a site using semantic HTML to aid accessibility
+- [ ] We can add appropriate attributes to HTML elements to make them screen reader friendly
+- [ ] We can add ARIA labels (where necessary) to make our semantic HTML more accessible
+- [ ] We can run a dev tools accessibility audit to check our website’s accessibility
```
arrested-developer: This duplicates requirements of the final LO (line 41), can this be removed?

learning-outcomes.md
```diff
@@ -0,0 +1,59 @@
...
+- [ ] We can use a screen reader to understand how people with visual impairments may experience our website
+- [ ] We can design a UI without relying solely on colour, so that we don't exclude colour-blind users
```
arrested-developer: These two (line 39 and 40) could be combined

- we can choose colours and contrast ratios in our UI to make our design inclusive for colour-blind and visually impaired users

learning-outcomes.md
```diff
@@ -0,0 +1,59 @@
...
+---
+
+### Version control
+
+- [ ] **We can use Git and GitHub as version control for our code**
+- [ ] We can initialise a Git repository so that we can track changes to our code over time
+- [ ] We can add / stage files so that we can include these updates in our next commit
+- [ ] We can write descriptive commit messages so that our commit history tells a story
+- [ ] We can amend commit messages so that we can correct typos/mistakes before we push to GitHub
+- [ ] We can co-author commit messages so that our commit history reflects everyone’s contributions
```
arrested-developer: is co-authoring necessary? We may be putting too much into their first week using Git collaboratively. I'd say this could be brought in in a later week

learning-outcomes.md
```diff
@@ -0,0 +1,59 @@
...
+
+---
+
+### Version control
+
+- [ ] **We can use Git and GitHub as version control for our code**
+- [ ] We can initialise a Git repository so that we can track changes to our code over time
+- [ ] We can add / stage files so that we can include these updates in our next commit
+- [ ] We can write descriptive commit messages so that our commit history tells a story
+- [ ] We can amend commit messages so that we can correct typos/mistakes before we push to GitHub
```
arrested-developer: Agree that amending commits is good to learn, but not sure if it's overloading them in the first week...

oliverjam: Yeah we can probably spread git stuff out a bit more across the 8 weeks

learning-outcomes.md
oliverjam: If we're going to mention any specific access requirements I would quite like to keep these as independent points, as they're not really related (other than being colour).

I do wonder if it makes sense to have a random selection of a11y stuff in here though. There's a big checklist of stuff to do/not do, and it feels weird to pick out just a few of them (and might mislead students into thinking these are more important). Maybe we should have a more generic a11y LO?

learning-outcomes.md
bobbysebolao: It probably isn't necessary in week 1, I put it here because it's something that FAC17 learned in their final week that they wish they'd learned earlier. It probably is a bit much for week 1 though. I think we can put this one in the toolkit set of LOs for a subsequent week

bobbysebolao:
I've attempted to address all of the suggested changes. @oliverjam I agree that there is an issue with having a "random selection of a11y stuff" in week 1 - I think we should have accessibility LOs for each week. I can amend weeks 2-4 LOs accordingly. 

I also removed some of the more advanced git concepts from week 1. I will add them to the Developer Toolkit LOs for a subsequent week.

Seems like this might be about ready for merging?

oliverjam:
I'm happy with the LOs. You seem to have 2 copies of them though, one inside the folder and one without. Also a `.DS_Store` snuck in

PR #16, opened 2020-01-22
Add testing week schedule

oliverjam:


week-3/schedule.md
```diff
@@ -0,0 +1,55 @@
...
+| _13:00_ | _Lunch_                 |                                        |
+| 14:00   | Project intro           |                                        |
+| 14:15   | Research                |                                        |
+| 16:30   | Research presentations  |                                        |
+
+## Day 3
+
+| Time    | Activity            | Learning outcomes           |
+| ------- | ------------------- | --------------------------- |
+| 10:00   | DOM Manipulation MC | add/modify/remove DOM nodes |
```
oliverjam: I think it's a good idea for the Wednesday morning Morning Challenge to be relevant to the project they're about to do. i.e. here they're about to build a to-do list with lots of DOM manipulation, so lets get good practices in their head right before they start.

oliverjam:
@sofer does it make sense on project days? There's no workshop for the reading to lead into, and it's less important to start exactly at 10.

sofer:
I still think it's worth making the start time 9:45 and not 10:00 on all 5 days.

oliverjam:
Sure, I can change the other days

oliverjam:
@sofer Done

PR #20, opened 2020-02-07
Create Week 2 schedule

oliverjam:
There are quite a lot of missing resources for this week that should be prioritised. I've got a decent idea of what I want them to look like though.

Also I moved the DOM manipulation morning challenge from Week 3 to this week because it's in the learning outcomes and is better to have as early in the course as possible. I've replaced it in Week 3 with a yet-to-be-created "Function scope & closures" MC, which fits better with Week 3's learning outcomes anyway.

### DIFF
--- a/week-2/schedule.md
+++ b/week-2/schedule.md
@@ -17,19 +17,21 @@
 
 ## Day 2
 
-| Time    | Activity                          | Learning outcomes |
-| ------- | --------------------------------- | ----------------- |
-| 09:45   | Intro reading                     |                   |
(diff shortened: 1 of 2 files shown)


sofer:
@oliverjam Have we got space for the second half of the issue management talk on Tuesday afternoon as part of the project intro (perhaps an extra 15 minutes would be enough if the project intro is kept short)?
- https://hackmd.io/@sofer/B1AL4V3ML#/

oliverjam:
yeah we can definitely get this in there

oliverjam:
@sofer done

PR #22, opened 2020-02-07
Add Week 8 schedule

oliverjam:
I've moved mocking from Week 7 MC to here since it's a better fit (they should be testing this frontend without worrying about the separate backend, which should already have tests)

Since this week is mostly stuff they already know the basics of I figured we could teach some more interesting client-side stuff. So I've added both ES Modules and basic client-side routing, both of which will make their projects easier and more realistic.

Let me know if you think that's too much new/weird stuff!

### DIFF
--- a/week-8/schedule.md
+++ b/week-8/schedule.md
@@ -1,0 +1,60 @@
+# Full stack app Week Schedule
+
+## Day 1
+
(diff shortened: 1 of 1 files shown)


bobbysebolao:
This looks so awesome!!!!

PR #23, opened 2020-02-09
Create week 5 schedule

bobbysebolao:
I've added the UI components/elements design burst to go with the design learning outcome (UI components) that we agreed for this week. 

The technical spike afternoon is shorter than usual, so it's perhaps worth finding some quite specific topics for students to spike on.

Otherwise, everything is similar to the current db week.

### DIFF
--- a/week-5/project.md
+++ b/week-5/project.md
@@ -1,0 +1,35 @@
+# The Project
+
+This week's project will involve setting up a database which you connect to via a node.js server. You'll use your data to make a dynamic web app for your front-end.
+
+Some suggested project ideas are below. Feel free to modify according to your interest, provided your idea has similar functionality.
+
+We're expecting to see:
+
+- A simple web app with a node server and a database
+- A schema for your database, which should be documented in your readme (along with any other architectural decisions)
+- Your database hosted on Heroku, or locally
+- A build script for your database
+- Security concerns appropriately considered (you must protect against script injections!)
+- Dynamic content, but DOM manipulation kept to a minimum
+- Mobile-first design
+- Clear user journey (even if you take one of our suggested ideas, document the user journey in your readme)
+- Test your server routes with supertest
+- Test your pure functions both server and client side
+- Set up a test database so that you can test your database queries
+
+**Note** We don't expect you to authenticate users (i.e. have a login or signup page), or even to simulate this feature. We'll cover how to do that properly in next week. Since these ideas were designed with Founders & Coders users in mind, we'll rely on trust instead of authentication :)
+
+## Suggested ideas
+
+- Founders & Coders book sharing system
+- Food / coffee recommendations around Founders & Coders
+- Founders & Coders events calendar
+
+## Getting started
+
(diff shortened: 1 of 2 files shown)


week-5/schedule.md
```diff
@@ -0,0 +1,58 @@
+# Database Week Schedule
+
+## Day 1
+
+| Time    | Activity                                                                                                                                                                                      | Learning outcomes |
+| ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------- |
+| 09:45   | Intro reading                                                                                                                                                                                 |                   |
+| 10:00   | [Intro presentation](https://docs.google.com/presentation/d/14LXEKmHM6xqjTvPLyKw0trtprGeNkD0VLgWkE8Z2ouo/edit#slide=id.g4dfce81f19_0_45)                                                      |                   |
+| 11:00   | [PSQL installation](https://github.com/macintoshhelper/learn-sql/blob/master/postgresql/setup.md) and [SQL commands intro workshop](https://github.com/foundersandcoders/sql-commands-intro/) | PSQL, SQL queries |
```
oliverjam: Given the intro presentation shouldn't take more than 30 mins, I structured the other weeks to have the intro/installation stuff from 10:30-11:00. It doesn't really change the actual schedule, but could we do that here for consistency with the other weeks?

bobbysebolao: yes will change that now

bobbysebolao:
@oliverjam good point, so I've moved the design burst presentation forward to 15:30 on day 1, which means the design burst mini workshop is at 10am on day 2. This makes the schedule slightly different to other weeks, but I think it's the best way to ensure we keep all the workshops so that people get as much practice writing SQL as possible.

week-5/project.md
```diff
@@ -0,0 +1,35 @@
...
+- Founders & Coders book sharing system
+- Food / coffee recommendations around Founders & Coders
+- Founders & Coders events calendar
+
+## Getting started
+
+Make sure you have a plan, and break the project down into manageable parts. Here are some things to consider:
+
+- You will need to make the requests and update the DOM in response using client-side JavaScript.
+- As well as serving static HTML and JS files, your server will also need to provide endpoints that return DB query results as JSON. You can query your server from the client using the fetch method.
```
oliverjam: I think we need to decide if weeks 4, 5 & 6 should be server-rendered or not. Currently I think we do a bad job of distinguishing, so some groups return HTML from their server and some return JSON. We should decide which way we want to go and be specific.

Given Week 8 is going to be "request JSON from your server and then render it client-side" maybe the Node weeks should be fully server-rendered? It will require us to have taught some rudimentary templating though. Could be something as simple as:

```js
function home(req, res) {
  getPostsFromDatabase().then(res => {
    const html = `
      <ul>
        ${res.rows[0].map(post => `<li>${post.title}</li>`)}
      </ul>
    `
  res.send(html)
}
```

PR #24, opened 2020-02-10
Create week 6 schedule

bobbysebolao:
Creates the schedule.md auth week schedule doc and the project.md project description doc.

**GAPS**
- There's no 30 min challenge to accompany the typography design burst presentation
- Research afternoon topics

### DIFF
--- a/week-6/schedule.md
+++ b/week-6/schedule.md
@@ -1,0 +1,58 @@
+# Authentication Week Schedule
+
+## Day 1
+
+| Time    | Activity                                                                                                                                                                                      | Learning outcomes |
+| ------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------- |
+| 09:45   | Intro reading                                                                                                                                                                                 |                   |
+| 10:00   | [Intro presentation](https://docs.google.com/presentation/d/14LXEKmHM6xqjTvPLyKw0trtprGeNkD0VLgWkE8Z2ouo/edit#slide=id.g4dfce81f19_0_45)                                                      |                   |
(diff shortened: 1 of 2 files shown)


week-6/schedule.md
```diff
@@ -0,0 +1,57 @@
...
+| 14:30   | Technical spikes                                                                                                                  |                                           |
+| 16:30   | Spike presentation prep                                                                                                           |                                           |
+| 17:00   | Spike presentations                                                                                                               |                                           |
+
+## Day 3
+
+| Time    | Activity                                                   | Learning outcomes   |
+| ------- | ---------------------------------------------------------- | ------------------- |
+| 09:45   | Intro reading                                              |                     |
+| 10:00   | [Promises MC](https://github.com/oliverjam/learn-promises) | Promises, fetch API |
```
oliverjam: Some form of this will be given in Week 2 now, since they need to learn promises to use `fetch`. We should find a different/more relevant to the project MC for this week

bobbysebolao: Good point. How about a new challenge where you have to refactor some bcrypt functions that hash passwords/compare a user's input with a stored hash to use promises instead of callbacks? All we'd have to do is give them the calback solution from the [day 1 password management workshop](https://github.com/foundersandcoders/ws-password-management) and get them to refactor it. What do you think?

Albadylic: Could we move the Client Side Validation into this slot and then add a typography workshop to space after the design burst?

oliverjam: @bobbysebolao I was planning on just teaching `bcrypt` and `pg`'s promise APIs from the start, since they'll have been doing promises the whole time.

@Albadylic I think client-side validation is a good fit actually 👍

PR #61, opened 2020-02-13
Add First-class Functions workshop

### LINKED ISSUE #33: First-class functions workshop
Explain the concept of functions being  "first-class" in JS. Lay the groundwork for callbacks. https://oliverjam.es/blog/first-class-functions/

oliverjam:
https://github.com/oliverjam/first-class-functions

Same style as the last one, which I'm enjoying (putting little check-in "challenges" as they read, then one harder "workshop" at the end)

I kinda want to get something like [this example](https://github.com/foundersandcoders/master-reference/issues/433) in there too, not sure.

Closes #33


bobbysebolao:
Looks great, I like the mini challenges interspersed throughout 👌

PR #63, opened 2020-02-14
Add HTTP intro slides

### LINKED ISSUE #31: HTTP Intro
Slides and mini workshop covering how the web works, HTTP requests, methods, bodies. Maybe demo `curl`?

https://hackmd.io/FPs_TDZIT9-dFOWtpJNUkQ?both

oliverjam:
Not quite polished yet but getting there

https://hackmd.io/@oli/Hy2LcbNm8

Closes #31

### DIFF
--- a/week-2/schedule.md
+++ b/week-2/schedule.md
@@ -5,13 +5,14 @@
 | Time    | Activity                              | Learning outcomes    |
 | ------- | ------------------------------------- | -------------------- |
 | 09:45   | Intro reading                         |                      |
-| 10:00   | Intro presentation                    |                      |
+| 10:00   | [Intro presentation][http-slides]     |                      |
 | 10:30   | [HTTP intro][http-intro]              | HTTP                 |
 | 11:00   | [Fetch & Promises intro][fetch-intro] | Fetch, Promises      |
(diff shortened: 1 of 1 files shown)


bobbysebolao:
Looks great @oliverjam ! My only suggestion is that we should think about including more visual content in the slides later on down the line, to have multiple ways of communicating the same info.

Albadylic:
These are good slides - they're clear and I think could be delivered by mentors easily. I think it offers some opportunity for discussion and knowledge sharing amongst the cohort. 

One thing I'd suggest changing is the information on layers. If they only need to know about the application layer - it might not be worth mentioning the other layers. In my experience, students find it frustrating to be told something but that they don't need to worry about it.

However, if you think the other layers provide useful context - my suggestion would be to alter the top paragraph on application layer to something along the lines of "The most important one for us, is the application layer". 

This set of slides is definitely a good resource for us to look at as precedent for each week. :star2:

oliverjam:
@Albadylic Thanks, good feedback. I've consolidated the 4 layers onto a single slide to de-emphasise their importance, and changed the Application Layer slide to use your suggestion.

PR #65, opened 2020-02-17
Add week 2 "Real world fetch" workshop

### LINKED ISSUE #35: Submitting data with fetch
How to `POST` data with fetch.

- URLencoded & JSON
- Different ways to access form values
    1. Directly accessing `querySelector("#my-input").value`
    1. Using `event.target.elements[name].value`
    1. `Object.fromEntries(new FormData(form))`

oliverjam:
https://github.com/oliverjam/real-world-fetch

- Covers stuff you need to use `fetch` in a real app
    - Error handling for non-200 HTTP responses
    - `fetch` options:
        - method
        - body
- Handling user input with forms
    - `preventDefault()`
    - Various ways to access form input data

I'm not sure if it needs a proper "workshop" bit at the end. Some of the interspersed "challenges" are more involved and are basically what I'd want from a "workshop" anyway.

I also swapped the first-class functions and learn-fetch-promises workshops around, since we should cover callbacks before promises.

Closes #35

### DIFF
--- a/week-2/schedule.md
+++ b/week-2/schedule.md
@@ -18,20 +18,21 @@
 
 ## Day 2
 
-| Time    | Activity                                           | Learning outcomes                 |
-| ------- | -------------------------------------------------- | --------------------------------- |
-| 09:45   | Intro reading                                      |                                   |
-| 10:00   | [Design burst: layout][db-layout]                  |                                   |
-| 10:30   | Layout workshop (tbc)                              | CSS layout                        |
-| 11:00   | Submitting data with Fetch                         | POST requests                     |
-| _13:00_ | _Lunch_                                            |                                   |
-| 14:00   | [Issue management & estimation][estimation-slides] | Project management, GitHub issues |
-| 14:15   | Project intro                                      |                                   |
-| 14:30   | Technical spikes                                   |                                   |
-| 16:30   | Spike presentation prep                            |                                   |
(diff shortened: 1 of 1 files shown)


bobbysebolao:
I ran through this and I think it's really good. I think you're right that the challenges throughout are involved enough to make this feel like a workshop with enough chances for students to practice the concepts that are covered for themselves 👍

Albadylic:
p.s. I definitely learnt a couple of things from reading this!

oliverjam:
@Albadylic cheers, fixed that bit. I think it could still do with a "workshop" part at the end because this is supposed to fill 2 hours. It's currently a 5 or 6 minute read (obviously longer with all the breaks to do the code challenges/let stuff sink in). It would be nice for people who have done some fetch before and whizz through the simpler parts to be able to push themselves a bit.

Albadylic:
@oliverjam Yeah, I reckon this would benefit from a workshop section to fill the two hours. It could be very similar to the challenges, as in the learn-fetch workshop, but on a new api - something that is likely to 404.

oliverjam:
@Albadylic I've added a simple-ish workshop at the end using the same PokéAPI they used in the first fetch workshop. This time they actually have to make a form and render to the page, plus handle errors.

Albadylic:
Anytime you use the PokeAPI, you get my automatic approval.

PR #84, opened 2020-03-03
Add new week 1 spikes

oliverjam:


bobbysebolao:
Yeah, I agree these topics look like a good fit for week 1 👍

PR #93, opened 2020-03-09
Add week 2 spike topics

oliverjam:


oliverjam:
They should hopefully learn what it means by reading the linked articles/doing the research. Maybe I should rephrase to avoid it

oliverjam:
I just removed that bit, it works without it

PR #96, opened 2020-03-12
Add week 3 project guide

oliverjam:
How do we feel about providing the test shells for them to fill in?


oliverjam:
@Albadylic Yeah they should definitely be writing more tests than these. They'll also get a fair bit of practice on Mon/Tue so I'm not too worried

PR #104, opened 2020-03-31
Add Precouse Doc

Albadylic:


oliverjam:
Do we definitely want to direct applicants to the coursebook before they start? We kind of did an official introduction in-person on Monday Week 1 for FAC19

Albadylic:
That's a good point - I added this here as per Dan's suggestion on the Master Ref version of this doc. We could make a precouse GitBook? Although that's a whole new box of frogs to open 🐸

oliverjam:
I guess the current precourse is in the Master Reference. The only thing stopping them from looking at the curriculum from there is the general disarray of that repo 😅

oliverjam:
I tried to merge this while github was having server issues last night, so it looks like the commit was merged but the PR didn't close. I'll do it manually

PR #107, opened 2020-04-03
Fix a subhead markup glitch

sofer:
Strange.


sofer:
It was a character that looked like a space, but wasn't.

oliverjam:
Ugh this happens to me a lot. You have to hold `alt` to type a `#` on a Mac. Unfortunately `alt-space` types a non-breaking space (which breaks markdown formatting), so if I don't let go of `alt` fast enough after the hash this happens.

I need to configure my editor to treat `alt-space` as a regular space

Issue #121, opened 2020-04-15
React Week


sofer:
Add the README from the existing repo.

PR #130, opened 2020-04-22
Move React Week resources into coursebook

oliverjam:
- [x] Learning outcomes
- [x] Schedule
- [x] Project guide


sofer:
Minor point: I think I got it wrong in #121. It might be better to describe this as Week 9, which is where in future it is very likely to go. I'll raise a separate issue for it, so as not to hold up merging this.

oliverjam:
I'll open issues for both of those changes

PR #145, opened 2020-05-04
Design week

Albadylic:
- Add schedule for Design Week
- Add docs for learning outcomes and project


design-week/learning-outcomes.md
```diff
@@ -0,0 +1,29 @@
+# Design Week Learning Outcomes
+
+### UX and Design
```
oliverjam, line 3: Not a blocker but this should be an `h2` (unless there's a missing heading that should be above it)

```suggestion
## UX and Design
```

Issue #122, opened 2020-04-15
Mentor guidance for the second half of the course


sofer:
In particular, we want a project reviewer who could do some live refactoring of project code in each of the one-week build sprints.

Also, all four role mentors.

sofer:
[What's the script?](https://github.com/foundersandcoders/london-programme/issues/1025) is a good starting point. Perhaps we can just move that guidance into the coursebook.

sofer:
In particular, we need pretty good guidance for any mentors doing refactoring. @oliverjam has offered to review the first draft.

oliverjam:
### Circle discussion

Create a `projects` (or something) folder in the Coursebook with second half of the course guidance.

sofer:
@Albadylic to make a pr in this repo for the [mentor guidance](https://github.com/foundersandcoders/london-programme/issues/1025#issuecomment-620511704).

sofer:
@Albadylic to look at this tomorrow and post pr.

PR #146, opened 2020-05-06
Projects Section

Albadylic:
Relates #122


Albadylic:
Think everything should be here now - have added guidance for mentors and students for project roles and a schedule for weeks 12,13,15,16. We may want to add LO's in the future.

PR #169, opened 2020-06-15
Add guidance for others using our curriculum

### LINKED ISSUE #168: Add guidelines on how to use our open source curriculum
It seems the existing open source curriculum with a CC license is not explicit enough, so let's add this: 

- [ ] What you can do
- [ ] What you cannot do 
- [ ] Human readable version of https://creativecommons.org/licenses/by-nc-sa/4.0/

oliverjam:
We should make it clear that we'd love for others to use our curriculum as a learning resource. However we also need to explain exactly how our license works so there is no confusion

Fixes #168

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ The work is available under the Creative Commons [Attribution-NonCommercial-Shar
 
 1. Give us credit for creating it
 1. Maintain the original license
-1. Do not use it in a commercial project.
+1. Do not use it in a commercial project
 
 #### What you can do
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Founders and Coders Curriculum
 
-Welcome to our web development curriculum. This is the material we use to teach beginners to be developers in our free bootcamp.
+Welcome to our curriculum. This material forms the basis of our tuition-free, peer-led web development programme.
 
 ## Using our curriculum
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,32 @@
-# coursebook
+# Founders and Coders Curriculum
 
-The new coursebook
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,6 +1,32 @@
-# coursebook
+# Founders and Coders Curriculum
 
-The new coursebook
+Welcome to our web development curriculum. This is the material we use to teach beginners to be developers in our free bootcamp.
```
Albadylic, line 3: Could we avoid using the word 'teach'? I'm not sure it reflects the learning environment we create. I could suggest: "This the the material which forms the basis of our tuition-free, peer-led web development programme."

README.md
```diff
@@ -1,6 +1,32 @@
...
+
+If you do decide to use our material please let us know. We can only accept so many people onto our program so we'd love to hear about others benefiting from our work. You could also join our Slack community to help and be helped by other people learning to code.
+
+### License
+
+The work is available under the Creative Commons [Attribution-NonCommercial-ShareAlike](https://creativecommons.org/licenses/by-nc-sa/4.0/) license. This means you can share or adapt the material as you like, as long as you:
+
+1. Give us credit for creating it
+1. Maintain the original license
+1. Do not use it in a commercial project.
```
Albadylic, line 17: Full stop on line 17 could be removed

README.md
oliverjam: Good catch, thanks

README.md
yvonne-liu: Do we think "do not use it in a commercial project" is explicit enough so people understand they can't take our curriculum and charge other people to use it?

oliverjam: Good point. I will make that clearer—"You cannot charge for anything using this material"

PR #174, opened 2020-06-18
Add presentation guidance

Albadylic:
Still unsure of the links - have added them to Summary.md, this is the only difference I can see between the routing here compared to Precourse.md 🤷‍♀️ 

Have added in some guidance for TFB presentations

### DIFF
--- a/projects/tfb-presentation.md
+++ b/projects/tfb-presentation.md
@@ -56,10 +56,10 @@ The purpose of this document is to provide advice about how to prepare a Tech fo
 
 ### Handy Tools
 
-[Excalidraw](https://excalidraw.com/) - make "handdrawn" mock ups of your UI
-[Creately](https://creately.com/) - create process diagrams (i.e. to map the flow of your app)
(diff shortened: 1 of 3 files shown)


projects/tfb-presentation.md
```diff
@@ -0,0 +1,65 @@
...
+
+---
+
+### Handy Tools
+
+[Excalidraw](https://excalidraw.com/) - make "handdrawn" mock ups of your UI
+[Creately](https://creately.com/) - create process diagrams (i.e. to map the flow of your app)
+[Dbdiagram.io](https://dbdiagram.io/home) - visualise your db schema
+[Gyazo](https://gyazo.com/captures) - record GIFs of your screen
+[Canva](https://www.canva.com/) - professional presentation templates
```
oliverjam, line 63: These either need to be a list or have an additional line break (so the markdown engine puts each in a new paragraph). Right now they are all output on a single line in the preview

Albadylic: Good spot, I copied from a HackMD 😅

Issue #177, opened 2020-06-20
A new default schedule


sofer:
Address #164 and introduce some other changes that will need further discussion:

1. **Shorten the Monday intro from 30 to 15 minutes** My impression is that once we go much beyond 20 slides it becomes difficult to take much more in. Keeping the intro short might discourage presenting too much info. On the other hand, I do like Oli's slides, so maybe we can compromise on 20 mins and re-write the other intros;
2. **45-min intro workshop on Monday** Most weeks seem to have an intro workshop of 30 or 60 minutes, so separate this out as a single 45-minute workshop;
3. **All other workshops are 2 hours long** Most of them already appear to be, but make this a general rule of thumb. There may be some workshops that need more time, but it is probably easier for planning new workshops and schedules if we aim for all of them to be a similar length;
4. **15-min project intro Monday pm** which is probably where it belongs, right before the spikes;
5. **Spikes across Monday and Tuesday** which is where they settled for FAC19;
6. **Tech for Better and Speaker** moved to a 2-hour slot at the end of the day on Tuesday now that the TfB slot is a conversation, not a workshop;
7. **Employment prep and community engagement** moved to a 2.5-hour slot at the end of the day on Wednesday, which if we don't use up every week can be given back to project time;
8. **Friday presentation time shortened to 90 minutes**. Over-long presentations in the first half of the programme have been a feature of FAC for as long as I can remember. This has to stop. The extra time can go to SGC (or a tea break);
9. **Added a 5-minute checkout**. 15 minutes is too long for the check-out and it never starts on time.

sofer:
Friday:
14:00 Presentations
15:30 SGC
16:15 Team SGC
17:00 Speaker

sofer:
Mon:
10:00 Mentor intro
10:05 Week intro
10:20: intro project

sofer:
Merged.

PR #178, opened 2020-06-20
Update Week 1-9 default schedule

sofer:
Addresses #177, which is still awaiting discussion.


README.md
```diff
@@ -40,59 +40,66 @@ The work is available under the Creative Commons [Attribution-NonCommercial-Shar
 | 6    | [Authentication](./week-6/schedule.md)       |
 | 7    | [REST APIs](./week-7/schedule.md)            |
 | 8    | [Full-Stack App](./week-8/schedule.md)       |
```
oliverjam, line 42: Should we insert Reading Week here as a break? Technically Full-Stack App is Week 9 and React Week is Week 10 right?

README.md
```diff
@@ -104,7 +111,9 @@ The general timetable for the first 8 weeks. This does differ from week to week
 | 12:15   | Presentation Prep   |                                                                            |
 | _13:00_ | _Lunch_             |                                                                            |
 | 14:00   | Presentations       | Present your project from this week                                        |
-| 16:00   | Stop, Go, Continues | Cohort and team retrospectives                                             |
-| 16:45   | Update User Manuals | Update user manuals with reflections from the week                         |
-| 17:00   | Speaker             | A talk about careers and employment from an employment partner or FAC alum |
+| 15:30   | Stop, Go, Continue  | Cohort retrospective                                                       |
+| 16:15   | Team retrospectives | SGC and update user manuals                                                |
+| 17:45   | Curriculum feedback | Online survey                                                              |
```
Albadylic, line 116: 16:45

README.md
sofer: Interesting question. I am not sure if reading week is going to stay as Week 8. It could shift a week or so in either direction, no? Also, 1 cohort out of 3 it's two weeks. Maybe we should just ignore it for now and treat the first part of the programme as a 9-week unit, regardless of the breaks.

oliverjam: Fair enough. I just want to make sure we don't confuse people, since planning that break in advance will be important once they're allowed to actually go do fun stuff like holidays again

PR #184, opened 2020-06-29
Add schedule for week 18

Albadylic:
Added schedule for Week 18 including
- This week's talks
- An hour to come up with a mentoring CoC - I am calling "Mentoring expectations"
- An hour for each pair to give curriculum feedback (adding issues to GitHub?)

### DIFF
--- a/SUMMARY.md
+++ b/SUMMARY.md
@@ -108,3 +108,6 @@
 
 - [Schedule](week-17/schedule.md)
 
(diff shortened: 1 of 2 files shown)


sofer:
Quick question: Do these even belong in the coursebook? I wonder if one-off schedules would be better off in london-programme.

Albadylic:
That's a good point, the next CF would have to rewrite this file in 4 months.

PR #196, opened 2020-07-10
Fix broken topic links in readme

### LINKED ISSUE #195: Readme links are broken
Hi!

Cool knew coursebook! 👀 

Great job everyone involved

Tiny thing I thought you should know that at the topics section the week links appear broken:

https://github.com/foundersandcoders/coursebook#topics

I guess due to #188 

If I get a chance I'll make a PR but thought I'd flag in case someone else gets a chance before me (also not sure of the correct process) :D

oliverjam:
Fix #195


oliverjam:
Good catch—can we raise a separate issue/PR for that change?

PR #202, opened 2020-07-16
timings and update to week 2 structure

jackherizsmith:
- updated day end timings in default to 17:45
- updated Friday timings to accommodate SGCs and user manual updates
- fixed a link in week 2
- shuffled week 2 (t4b and spikes) with fetch workshop slightly spanning lunch since the fetch workshop is juicy.
- updated links to workshop lengths

### DIFF
--- a/curriculum/http/schedule.md
+++ b/curriculum/http/schedule.md
@@ -2,53 +2,61 @@
 
 ## Day 1
 
-| Time    | Activity                              | Learning outcomes    |
-| ------- | ------------------------------------- | -------------------- |
-| 09:45   | Check-in                              |                      |
-| 10:00   | [Intro presentation][http-slides]     |                      |
-| 10:30   | [HTTP intro][http-intro]              | HTTP                 |
-| 11:00   | [First-class functions][fc-fns]       | Functions, callbacks |
-| _13:00_ | _Lunch_                               |                      |
-| 14:00   | [Fetch & Promises intro][fetch-intro] | Fetch, Promises      |
-| 16:00   | Tech for Better                       |                      |
+| Time    | Activity                                           | Learning outcomes    |
+| ------- | -------------------------------------------------- | -------------------- |
+| 09:45   | Mentors intro                                      |                      |
+| 09:50   | Check-in                                           |                      |
+| 10:00   | Feedback survey                                    |                      |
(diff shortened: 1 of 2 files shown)


oliverjam:
I _think_ this is cool, although I'm not sure about having 30 mins on a workshop before a lunch break. Might be a little disruptive. Did you also make the design burst 15 mins longer?

curriculum/http/schedule.md
```diff
@@ -2,53 +2,61 @@
...
-| _13:00_ | _Lunch_                               |                      |
-| 14:00   | [Fetch & Promises intro][fetch-intro] | Fetch, Promises      |
-| 16:00   | Tech for Better                       |                      |
+| Time    | Activity                                           | Learning outcomes    |
+| ------- | -------------------------------------------------- | -------------------- |
+| 09:45   | Mentors intro                                      |                      |
+| 09:50   | Check-in                                           |                      |
+| 10:00   | Feedback survey                                    |                      |
+| 10:05   | [Intro presentation][http-slides-25]               | What is HTTP         |
+| 10:30   | HTTP game                                          | How does HTTP work?  |
```
Albadylic, line 11: I think we said we'd remove this - would we like to remove from schedule?

jackherizsmith: We're going to try and make HTTP game work

jackherizsmith:
Sorry for not replying sooner. There is a gap on Tuesday morning after spikes that I wasn't sure how to fill, and I remember that workshop being quite challenging and it has the a lot of the focus for the week in it, so I am essentially filling in the gap with a longer DB and earlier workshop. 

Could you look at the schedule and suggest another way around it? I could definitely have missed a better solution.

oliverjam:
So in later weeks we resolved this by having the employment prep hour (which is what replaced Design Bursts) in the afternoon. We took half an hour from the afternoon workshop and half an hour from TFB to make this happen (so 14:00-15:30 workshop, 15:30-16:30 employment, 16:30 TFB).

I'm not sure it makes sense to shift DBs to after lunch though. Maybe this structure is fine—having time to read through the workshop and then take a break before you actually get stuck in to coding.

jackherizsmith:
And does a longer DB make sense? I'll explain to H&J in advance the situation. Will layout work for 45 mins?

oliverjam:
Sorry I didn't see this—yeah that's all good

PR #244, opened 2020-08-03
add alternative links for spikes

jackherizsmith:


jackherizsmith:
Or the Markdown you posted?

oliverjam:
I would rather keep the official Node resources if possible. They appear to be having some kind of cache issue with their site right now, but it's not like the website has been closed down

PR #245, opened 2020-08-06
Clarify Node project shouldn't have client-side JS

oliverjam:


oliverjam:
I don't know if that's going too far/too prescriptive. There's no reason they can't have static HTML files for pages that never change. Maybe it would be simpler for them to think entirely in terms of templating for now though, I don't know

jackherizsmith:
I do think preventing people from making html files forces them to use this new approach, without wasting time on working out how to do it in the way they're more familiar with...

oliverjam:
😆 I guess enforcing it would be helpful in that case, I'll push a change

Maybe the flow of this week is broken since the [very first workshop](https://github.com/oliverjam/node-http-server) gives you all the info you need for the project. Everything after that isn't really necessary, but I imagine people forget what they did Monday morning by Thursday

PR #249, opened 2020-08-11
Swap old DB MC for new one

oliverjam:
We never actually did the old one for FAC19, I just never got time to replace it. The new one covers deploying Postgres to Heroku, then some fun advanced SQL stuff to practice joins and sub queries.


oliverjam:
Apologies for the unnecessary diff, looks like this file didn't get Prettier-ed last time it was edited, so mine made some formatting fixes to the tables.

PR #266, opened 2020-08-26
add fac19 react

jackherizsmith:


oliverjam:
@jackherizsmith I think we wanted to add new examples _in addition_ to the existing ones 😅 

It's not a big deal though, just the references to tamagotchis might be confusing

jackherizsmith:
Whoops sorry I'll add them back in!

PR #269, opened 2020-08-27
re-add examples

jackherizsmith:


curriculum/react/project.md
```diff
@@ -23,8 +23,13 @@ The goal is to build an interactive game-like thing that uses data from an API.
 "interactive game-like thing" is a bit vague, so here are two example apps:
```
oliverjam, line 23:
```suggestion
"interactive game-like thing" is a bit vague, so here are some example apps:
```

oliverjam (changes requested):
Just one reference to "two examples" to change

PR #278, opened 2020-09-07
Add SPA week day 2 workshops

oliverjam:
Plus half hour for student projects


jackherizsmith:
Whoops just saw this thanks Oli (and thanks for adding the time in)

PR #282, opened 2020-09-14
React week resources

oliverjam:
Not sure why I didn't copy the resources over with the rest of the week's content ages ago.

Also we need to use my fork of Sofia's workshop as there are some fixes in a PR that isn't merged yet.


jackherizsmith (approved):
looks good

PR #299, opened 2020-10-02
Updates to schedule

Albadylic:
Jack and Gregor to review on Monday

### DIFF
--- a/projects/tech-for-better/build-1.md
+++ b/projects/tech-for-better/build-1.md
@@ -5,7 +5,8 @@
 | Time    | Activity                     |
 | ------- | ---------------------------- |
 | 09:45   | Check-in                     |
-| 10:00   | Reach out to Project Mentors & Role mentors |
+| 10:00 | Feedback Survey |
+| 10:10   | Reach out to Project Mentors & Role mentors |
 | 10:30   | Build                        |
 | _13:00_ | _Lunch_                      |
 | 14:00   | Build       |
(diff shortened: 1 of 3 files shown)


projects/tech-for-better/build-1.md
```diff
@@ -5,7 +5,8 @@
 | Time    | Activity                     |
 | ------- | ---------------------------- |
 | 09:45   | Check-in                     |
-| 10:00   | Reach out to Project Mentors & Role mentors |
+| 10:00 | Feedback Survey |
+| 10:10   | Reach out to Project Mentors & Role mentors |
```
jackherizsmith, line 9: I wasn't sure about this in SP build 2 so removed it - I'm not sure if we want to allocate a specific time or leave it to them to find each other as and when they need it?

jackherizsmith (commented):
We might need to put a bit  more thought into tech spikes - unless instead we said Thurs pm could be spent doing deployment? Get all the boring stuff out of the way ahead of build week 1?

projects/tech-for-better/build-1.md
Albadylic: Happy to remove

Albadylic:
> We might need to put a bit more thought into tech spikes - unless instead we said Thurs pm could be spent doing deployment? Get all the boring stuff out of the way ahead of build week 1?

Yes, we could call it project setup. I'd be curious to try Tech Spikes again with little direction and see how they go.

jackherizsmith:
I can't ask you to review your own PR apparently but I have made some more changes

Albadylic:
All good - cheers!

PR #310, opened 2020-10-26
First draft of debugging advice

oliverjam:
I starting writing a "how to ask for help guide" but it ended up getting too technical so I split the debugging part out into this

### DIFF
--- a/handbook/debugging.md
+++ b/handbook/debugging.md
@@ -50,6 +50,12 @@ Ideally this will return GitHub issues on the repo for the library you're using,
 
 Once you understand what's wrong you need to fix your code. If you found an easy solution to your problem this might be simple, but sometimes it will require experimentation.
 
+### Make sure your code is running
+
+Always check that the code you think is executing _really is_. It's surprisingly common to be refreshing the deployed production site wondering why your local changes aren't showing up.
+
+Make sure the function you're changing is actually getting called, and if there are any branches in logic (e.g. `if` statements) that you're in the one you expect. This can be as simple as including a `console.log("here!")`, or as complex as stepping through your code using your [browser's debugger](https://javascript.info/debugging-chrome).
+
 ### Change one thing
 
 The biggest mistake beginners make when trying to fix a problem is changing multiple things at once. As with scientific experiments you should only ever change one thing at a time. This ensures that you know exactly what change was the right one.
--- a/handbook/debugging.md
+++ b/handbook/debugging.md
@@ -14,8 +14,6 @@ It's easy for your first reaction to be "I suck at this, no one else has these p
 
 Make sure you stop and understand exactly what went wrong before you start trying to fix it. Otherwise you might go down an entirely wrong path, which will make it harder to figure out what the real original problem was later.
 
-### Read the question
-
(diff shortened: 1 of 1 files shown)


handbook/debugging.md
```diff
@@ -0,0 +1,71 @@
...
+
+The most important thing is to remember that getting stuck doesn't say anything about you or your skill as a developer.
+
+It's easy for your first reaction to be "I suck at this, no one else has these problems". _Everyone_ makes typos, forgets how things work or gets inscrutable errors in their terminal. The mark of a more experienced developer isn't avoiding errors entirely, it's reacting to them constructively.
+
+## Understanding the problem
+
+Make sure you stop and understand exactly what went wrong before you start trying to fix it. Otherwise you might go down an entirely wrong path, which will make it harder to figure out what the real original problem was later.
+
+### Read the question
```
sofer, line 17: "Read the question" makes sense in some contexts, but not in others. Perhaps just lose this subhead?

handbook/debugging.md
```diff
@@ -0,0 +1,71 @@
...
+
+JavaScript errors generally consist of an [error type](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors), an error message, then a stack trace. Let's look at an example error and break it down:
+
+```
+Uncaught TypeError: arr.join is not a function
+    at arrayToString (index.html:11)
+    at run (index.html:14)
+```
+
+The first part tells us that our code encountered a "type error". This means we tried to use a value in an impossible way. E.g. accessing a property on `undefined` instead of an object.
```
sofer, line 37: It may be worth adding parenthetically that this is a *very common* source of error in JS and may be worth eliminating from your research before you do anything else.

handbook/debugging.md
```diff
@@ -0,0 +1,71 @@
...
+
+### Google the error
+
+If you can't figure out what's wrong from the error message/looking at your code your next step should be to search for the _exact error text_ on Google. You can put a search term in quotation marks to force Google to show you exact matches.
+
+Learning to craft searches that return the most useful results is one of the most important skills to grow. As well as searching for the exact error message you can try including keywords for more context, like "JS", "npm", "React" etc. You can also limit your search to specific sites by adding their domain, e.g. "site:github.com".
+
+Ideally this will return GitHub issues on the repo for the library you're using, or Stack Overflow questions from people who had the exact same problem. Try to prioritise recent stuff, since the JS ecosystem moves very fast. Usually a 2 year old issue won't be relevant anymore.
+
```
sofer, line 50: Arguably, this all belongs in a separate "How to search online"  README. Separately, do we want to use "google" as a verb?

handbook/debugging.md
oliverjam: Yeah that was meant to be metaphorical 😅. As in "read the question" is something my teachers drilled into me at school that is applicable to a bunch of other stuff. Probably not super clear though so dropping it probably makes sense.

handbook/debugging.md
oliverjam: I think google was verbified a long time ago 😅 https://en.wikipedia.org/wiki/Google_(verb)

I agree that we could probably write a whole document on finding good stuff online. When I get round to writing it I can move this section out and link to it

oliverjam:
@sofer thanks for the feedback, made some amends and added a new section

PR #358, opened 2021-02-17
Update coaching guidance

### LINKED ISSUE #352: Update coaching guidance
Even though I thought I was a fan of roles-based groups, the [guidance we have](https://github.com/foundersandcoders/coursebook/blob/main/src/course/handbook/coaching.md) may be overly complicated. I have drafted [an update](https://www.notion.so/foundersandcoders/Learning-circles-ce61ef400ca9424c87e6130f760b6946), which is partly intro to a "learning circle" format and partly how we could implement them in coaching and comunity meetups.

sofer:
Fixes #352.


oliverjam (commented):
Looks good. Should we delete the old `coaching.md`?

PR #359, opened 2021-02-18
Precourse week 3: JS

oliverjam:
Adds a workshop of first-class functions, callbacks, and async JS. This effectively combines the first-class functions workshop from week 2 with part of the async intro in the learn fetch workshop. We can slot other stuff into those gaps.


sofer (approved):
Nice.

PR #372, opened 2021-03-06
Week 1 schedule

oliverjam:
- Add new Week 1 schedule
- Fix default schedule overriding logic. It now correctly removes whatever was in the default schedule slot if a week schedule has an entry that overlaps at all. E.g. if you schedule something for 13:30 it'll remove lunch for that day. The downside is you have to manually re-include default stuff if you are tweaking times a bit (e.g. "Respond To Issues on Friday here).

Screenshot 2021-03-06 at 20 35 55

### DIFF
--- a/src/_data/defaultSchedule.yml
+++ b/src/_data/defaultSchedule.yml
@@ -15,6 +15,7 @@ monday:
   - start: 17:15
     end: 17:45
     name: Spike preparation
+    url: /handbook/spikes/#what-makes-a-good-research-presentation
   - start: 17:45
     end: 18:00
     name: Check-out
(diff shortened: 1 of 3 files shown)


sofer:
https://deploy-preview-372--fac-coursebook.netlify.app/course/syllabus/teamwork/schedule/

Tuesday, 3.30-5.30 there's a bit of a clash.

(also the new-style schedule is showing up for Weeks 1 and 3, but not for other weeks)

oliverjam:
Ah good catch, must still be a flaw in the logic for overriding the default schedule :(

Yeah I added Week 3 as a test when I first made the new schedule format. This PR just adds Week 1. I'll do each of the others when we need them

oliverjam:
Ah we missed the case where the overriding entry starts before __and__ ends after the default one. Should be fixed now

PR #391, opened 2021-03-18
Weekly schedules

Albadylic:
Mostly just updating links for next week. Could you have a look at Tuesday afternoon and check everything is in the right order?

There is 90 minutes spare for the Grid workshop challenge whereas before I think we gave it 30 so there may be some wiggle room in there if we want to add something to the schedule or move anything around.


oliverjam (commented):
Apologies but I'm in the middle of doing the links to the workshops, since they're moving onto this site instead of my GitHub. I should have warned you.

It's probably fine to merge this anyway and I'll overwrite those as I go

oliverjam:
Except I didn't realise this was here and merged one of my own changes beforehand and caused a conflict, apologies!

oliverjam:
Okay reverted mine so we can merge yours first, all good

oliverjam (approved):

Albadylic:
> Apologies but I'm in the middle of doing the links to the workshops, since they're moving onto this site instead of my GitHub. I should have warned you.
> 
> It's probably fine to merge this anyway and I'll overwrite those as I go

No worries! I should have checked with you so sorry about that. I had a feeling you'd have this in hand already.

PR #405, opened 2021-04-03
New learning circles copy

### LINKED ISSUE #404: Learning circles
The [current page on learning circles](https://learn.foundersandcoders.com/course/handbook/learning-circles/) needs attention. I'll make a pr before the next curriculum circle. It also needs a related crib sheet and glossary. I'll make a start on those, too. I'll add notes on why I'm proposing the changes in this issue.

sofer:
Resolves #404

### DIFF
--- a/src/course/handbook/glossary.md
+++ b/src/course/handbook/glossary.md
@@ -3,15 +3,14 @@
 Term | Meaning
 --- | ---
 **argument** | some input to a function
-**attribute** | like a variable, but specifically it's a variable that is defined as part of an *object*
 **function** | a piece of code that takes one or more inputs and returns a value as an output
 **function definition** | The code that defines how a function should behave
 **function call** | The code that makes use of an existing function definition with a specific set of inputs
 **input** | same as *argument*
 **method** | same as a function, but specifically it is a function that is defined as part of an object
-**object** | some code that defines some values (*attributes*) and some functions (*methods*) that are mutually dependent on each other
+**object** | some code that groups a set of named values together
 **output** | same as *return value*
 **parameter** | same as *argument*
 **return value** | the value that is output by a  *function* when it finishes running
 **type** | any one of a number of different kinds of value, such as a *string* or an *integer*, an *array*, or a *function*
-**variable** | some value defined in a *function*
+**variable** | Some named element to which a value can be assigned and updated
--- a/src/course/handbook/glossary.md
+++ b/src/course/handbook/glossary.md
@@ -1,0 +1,17 @@
+# Glossary or terms
+
+Term | Meaning
+--- | ---
+**argument** | some input to a function
+**attribute** | like a variable, but specifically it's a variable that is defined as part of an *object*
+**function** | a piece of code that takes one or more inputs and returns a value as an output
+**function definition** | The code that defines how a function should behave
+**function call** | The code that makes use of an existing function definition with a specific set of inputs
+**input** | same as *argument*
+**method** | same as a function, but specifically it is a function that is defined as part of an object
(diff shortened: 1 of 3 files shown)


sofer:
Probably best to wait until we've had a chance to discuss at curriculum circle.

sofer:
The circles copy and questions for problem solving and glossary can be used together but I have split them out as they may each have value independently of each other.

src/course/handbook/glossary.md
```diff
@@ -0,0 +1,17 @@
+# Glossary or terms
+
+Term | Meaning
+--- | ---
+**argument** | some input to a function
+**attribute** | like a variable, but specifically it's a variable that is defined as part of an *object*
```
oliverjam, line 6: I'm not sure I've ever heard an object's keys referred to as attributes before. Will it be confusing to people when the language itself has things like `Object.keys()`?

src/course/handbook/glossary.md
```diff
@@ -0,0 +1,17 @@
...
+**function** | a piece of code that takes one or more inputs and returns a value as an output
+**function definition** | The code that defines how a function should behave
+**function call** | The code that makes use of an existing function definition with a specific set of inputs
+**input** | same as *argument*
+**method** | same as a function, but specifically it is a function that is defined as part of an object
+**object** | some code that defines some values (*attributes*) and some functions (*methods*) that are mutually dependent on each other
```
oliverjam, line 12: This is quite a specific definition of an object.

I think there's a problem for beginners in that JavaScript uses objects both as maps/records/dictionaries (whatever other languages decided to call that data structure) _and_ and a way to group related data/methods.

Beginners mostly encounter objects as a map data structure, and very rarely create their own objects with methods (since OOP is quite uncommon on the web nowadays). If they are creating the OOP-style object they'd probably use a `class`.

A more general (but possibly less useful) definition might be something like: "some code that groups a set of named values together"

src/course/handbook/glossary.md
```diff
@@ -0,0 +1,17 @@
...
+**output** | same as *return value*
+**parameter** | same as *argument*
+**return value** | the value that is output by a  *function* when it finishes running
+**type** | any one of a number of different kinds of value, such as a *string* or an *integer*, an *array*, or a *function*
+**variable** | some value defined in a *function*
```
oliverjam, line 17: Again this is quite specific—a variable can be defined anywhere in JS

oliverjam (commented):
I like the general concept. I am a little concerned that the Handbook is getting pretty long and unwieldy. Perhaps it's going to need some kind of nesting so we can group sub-pages.

My only concern with these resources is they are very "function-focused" and clearly aimed at someone early in the Coaching programme. This is probably fine, but currently the only people who look at the Handbook are the full-time cohort.

sofer:
> I like the general concept. I am a little concerned that the Handbook is getting pretty long and unwieldy. Perhaps it's going to need some kind of nesting so we can group sub-pages.

Agreed. The solution will depend a bit on how many more of these sorts of pages get added.

> My only concern with these resources is they are very "function-focused" and clearly aimed at someone early in the Coaching programme. This is probably fine, but currently the only people who look at the Handbook are the full-time cohort.

Learning circles are also used (with a minor adjustment) in the interview prep tutorials. I still haven't run enough coaching sessions to be sure that this is quite the right format yet, but we may want to use them in the pre-apprenticeship and, quite possibly, in community meetups.

oliverjam (approved):

sofer:
I ran last night's coaching session broadly following this format and the emphasis on proceeding in rounds and on sparking a dialogue through getting participants to ask each other questions worked very well.

sofer:
Merging!

PR #448, opened 2021-05-05
Update employment mentors to interview prep

Oliversw:
Just to make sure that we're being clear about what is happening this afternoon! Changing the link from an employment mentor talk to the link for code writing exercises


sofer:
My only thought about this is that the Notion document might at some point work better as a github repo.

oliverjam (approved):

Oliversw:
@sofer happy to do so, adding this link now was just a last minute thought

PR #460, opened 2021-05-21
Design week

Albadylic:
- Create new section for projects
- Added schedules for Design Week
- Added schedules for the two build sprints. These need a wee bit of revision (adding in role circles and blocking out the bank holiday)
- Linked to a couple of useful documents

There's a tiny issue where the topic is not being generated into the `h1`, I couldn't work out how to get this to render it might be to do with the way I'm trying to structure the sidebar... 🤔


oliverjam:
Yeah the Syllabus pages rely on a [data file dynamically generating the title](https://github.com/foundersandcoders/coursebook/blob/main/src/course/syllabus/syllabus.11tydata.js) based on the file path. The simplest way to get this working properly would be to move the project stuff inside the `syllabus` directory, rather than having it separate

Albadylic:
Nice, thanks - that makes sense. I can do that.

Alternatively, could we make a new data file that does something similar but with the projects path? What do you think makes more sense in terms of layout? Are the projects part of the core syllabus or distinct?

Issue #461, opened 2021-05-25
Tech Spikes in Design Week


Albadylic:
We have tech spikes scheduled on Thursday afternoon.

We didn't have any particular guidance in place for what to cover, I think in FAC20 we asked them to consult with their mentor on what to cover.

They'll do tech spikes in their role groups.

https://learn.foundersandcoders.com/course/syllabus/final-project/spikes/

Would we like to give some more specific guidance?

oliverjam:
I think this worked okay for FAC20, so I'm happy to leave it as is

oliverjam:
Circle discussion: we'll keep it roughly as is. Clarify that they can work in pairs if they want to investigate different things (but not alone). The DevOps spike should recommend looking into staging environments to fulfill #328

- DevOps: staging env (e.g. Vercel/Netlify already auto-deploy branch previews, but they'll need to set up a separate staging DB)
- UX: styling methodologies
- QA: testing in GitHub Action CI
- Scrum: 🤷 

@Albadylic to update the Coursebook document

Issue #458, opened 2021-05-17
Workshops in first project build sprint


Oliversw:
There are currently some workshops on routing in React on Monday of week 1 of the build sprint. Do we still want to use that time for workshops? 
https://github.com/foundersandcoders/coursebook/blob/main/src/course/projects/student-projects/build-1.md

oliverjam:
Huh I forgot I even wrote those 😅. Since Next.js handles routing for you I don't think we need them this time around.

Albadylic:
Resolved here:
https://github.com/foundersandcoders/coursebook/pull/462

We might like to give them a step by step guide to project set up. @oliverjam - I'm happy for you to update the PR or let me know what to add in then I can add another commit.

oliverjam:
I think merge this and I'll add something to Tue morning later. Should probably also include a bit on how to approach building frontend-first to resolve #406

PR #462, opened 2021-05-25
Restructure and tidy up

Albadylic:
Moving some files into the right directories and updating schedules for build weeks


src/course/syllabus/final-project/schedule.md
```diff
@@ -0,0 +1,7 @@
+# Final Project Schedules
```
oliverjam, line 1: Is this file actually linked to from anywhere?

Albadylic: It is auto-linked under the Schedule tab for design and each build week. I added this to resolve the 404 when clicking that tab.

Albadylic:
Fixes #461, fixes #458

PR #477, opened 2021-06-28
Fix broken images in CSS Layout challenge

oliverjam:
Hotlinking to weird Instagram URLs was always a mistake 😅

(each image has a 1px different size to force Unsplash to return a different one, otherwise you get a grid of the same dog)


Albadylic (approved):
Beautiful

PR #481, opened 2021-07-01
Week 4

Albadylic:
Trying to get ahead of the game 😅 

Most of the content this week was from Tech Spikes so I'm suggesting we follow the pattern more closely. We might want to explicitly group them up for these.

Otherwise, adding a couple for JavaScript30 projects, expanding their fortnightly project and the final EP modules.


src/course/syllabus/pre-app-3/schedule.md
```diff
@@ -0,0 +1,36 @@
...
+
+Work through Array Cardio 2 and build a Flex Panels Image Gallery.
+
+## Execute Program
+
+Continue working on _Modern JavaScript_, _Arrays_ and _Concurrency_.
+
+Start working on _Regular Expressions_ and _SQL_.
+
+You'll be working on all of the available exercises and we won't be assigning more in the coming weeks. Structure your time so you can balance working on all five at once.
```
oliverjam, line 11: Do we want to be prescriptive about this? I.e. some people might prefer to finish the JS ones before starting on SQL

Albadylic: I'm not sure to be honest. I guess it depends on each person's working style and I think these two are less of a priority than the other three we've assigned. 

We could make a suggestion that they work on one topic each day or be prescriptive that they do all the JS ones first. Another option might be to leave SQL off for this week and add it to a later week so they don't feel overburdened.

oliverjam: Maybe we can be hands off and just say something like "Structure your time so you finish all the modules by the end of the pre-apprenticeship"?

Albadylic:
- Finish up next week's content
- Add content for Weeks 5 & 6
- Split Teamwork project into two sections: build the agency site, add a requests form
- Introduce spikes, code review, PR's

A lot of the new material will need to be covered in the slides to complement what's being introduced here.

(Confusingly, the file numbers become one behind the week numbers since week two was a break)

PR #487, opened 2021-07-07
Add weeks 5 & 6

Albadylic:


src/course/syllabus/pre-app-5/project.md
```diff
@@ -0,0 +1,49 @@
...
+Use what you've learnt in this week's DOM workshops to code an efficient and robust solution.
+
+Your requests section should:
+
+- Allow users to leave comments requesting to work with your agency
+- Attach a visible user name to each request
+- Ask the user to give an email address and validate the email
+- Limit the length of requests to 149 characters
+
+### User Stories
```
oliverjam, line 14: I think you want to remove this duplicate

src/course/syllabus/pre-app-5/project.md
```diff
@@ -0,0 +1,49 @@
+Your project for this week is to continue working on your [digital agency website](../../pre-app-4/project).
+
+You'll add a way to get in touch if they are interested in working with you. When the user clicks submit on the form, their request should be added to a visible backlog on the page. The backlog does not need to be preserved when the user refreshes the page.
```
oliverjam, line 3: I like this idea—my only concern is that it doesn't easily break down into separate async tasks. It's just one big feature, so might be difficult for them to manage working on separately

Albadylic: That's a good point - I guess only one person would be practising the template learnings and they rely on each other to finish different parts. 

I wonder if adding two of these kinds of features would be worthwhile or result in doubling the workload. We could ask them to arrange to code the template part together.

PR #494, opened 2021-07-26
Update and rename learning-circles.md to coaching-groups.md

### LINKED ISSUE #483: Learning circles -> coaching groups
The setup for group learning is continuing to shift. There is a consensus forming around the idea that we like coaching in groups of three more than groups of four, which makes the current suggestions for [learning circles](https://github.com/foundersandcoders/coursebook/blob/main/src/course/handbook/learning-circles.md) look a bit out of date already. [The format that FAC22 are using](https://hackmd.io/@sofer/S1QwzHo3d#/6) on Thursday evenings may be more useful both for peer-led groups and groups that are led by a coach. Let's replace the current guidance to reflect more recent practice.

sofer:
Fixes #483.


oliverjam (approved):

Albadylic:
I think for this to be reflected in the left navigation, you'd need to update the text here from learning-circles to coaching-groups.

https://github.com/foundersandcoders/coursebook/blob/main/src/_includes/course.11ty.js#L35

oliverjam (approved):

PR #501, opened 2021-08-02
Fix links that reference source files

oliverjam:
Links need to point to final built HTML pages rather than `.md` content files. The easiest way is to use relative URLs to reference sibling pages (e.g. `thing.md` -> `../thing/`)


Albadylic (approved):
Thanks for updating these

PR #505, opened 2021-08-09
Add Week 9 (testing) to pre-apprenticeship

oliverjam:
Week 10 will mostly be TDD Array Methods, the morning challenge and the bulk of the project work.

Screenshot 2021-08-09 at 15 18 22

Screenshot 2021-08-09 at 15 19 02


Albadylic (approved):

Albadylic:
I love the addition of learning outcomes and think the structure looks sensible for the two weeks. My only concern is that we're assigning 6 hours of expected work this week so they'll have to be careful about how they balance this work with any tech tests.

oliverjam:
@Albadylic how much are we supposed to be setting? Previous weeks looked like 5-6 hours. The unit testing workshop _could_ be done in an hour if they don't faff about, and we could drop the recommended time for the project setup (since they're just writing some tests)

Albadylic:
> @Albadylic how much are we supposed to be setting? Previous weeks looked like 5-6 hours. The unit testing workshop _could_ be done in an hour if they don't faff about, and we could drop the recommended time for the project setup (since they're just writing some tests)

Strictly speaking, I think we're splitting the 10 hours between the two evenings so aiming for 5 hours. I think 6 is okay and as you say, it's likely they can complete the work in less time.

Issue #475, opened 2021-06-18
A short Web science curriculum, perhaps


sofer:
Anni has suggested developing a series of sessions, perhaps 6 x 1 hour, for FAC22. It might fit well into one morning a week in the latter half of the programme while they are working on projects.

oliverjam:
I think we could use a general "lecture slot" for giving the cohort information that doesn't directly relate to the week topic. We've discussed putting that first thing on Wed, before the morning challenge and projects.

sofer:
Provisionally looking at 10-12 30-minutes sessions on a Wednesday morning. @sofer to set up initial conversation with @Intersticia and @oliverjam.

oliverjam:
Here's a list of stuff I think would be useful to cover (much of it inspired by [Tom Scott's The Basics](https://www.youtube.com/playlist?list=PL96C35uN7xGLLeET0dOWaKHkAlPsrkcha) YouTube series.

- Computer electronics
   - Binary
   - Switches
   - Logic gates/boolean logic
   - Machine code
- Computer science
   - Algorithms
   - Big O/time & space complexity
   - Cryptograph/hashing/encryption
   - Text encoding: ASCII/UTF-8/UTF-16
   - Text compression
- World Wide Web
   - The internet protocol stack: MAC/IP/TCP/HTTP
   - URLs -> domains -> DNS -> IP addresses
   - HTTPS encryption
- Ethics
   - Consequences/impact (e.g. civil engineers learn about the [Hyatt Regency disaster](https://en.wikipedia.org/wiki/Hyatt_Regency_walkway_collapse))
   - Obvious flaws missed by homogenous teams
   - Building safe social platforms
   - Algorithmic bias in machine learning/AI
   - How to not build a global platform for political disinformation and genocide
   - Employment rights & unions
- Security
   - How hacking works
   - White hat vs black hat
   - Famous exploit examples (e.g. [Samy](https://en.wikipedia.org/wiki/Samy_(computer_worm)))
   - Basics of protecting your sites

Intersticia:
Hey there Oli,

How about you and I set up some times over the next few weeks to do some curriculum development together?  

I’m pretty good from 8 am your time most days except Wednesdays (when I can do after 10 am), Thursdays (before 11 am your time) and Fridays (after 10 am your time).

Let me know what might suit on your end.

Very much looking forward to this!

Anni

oliverjam:
Notes from my call with @Intersticia:

[...]

Issue #473, opened 2021-06-15
Bringing Tech for Better back into the programme


Albadylic:
Is there space to move the first 3 weeks of the curriculum into the pre-apprenticeships and running two projects at the end of the course.

Albadylic:
All agreed to nominally move the first 3 weeks of the course into the pre-apprenticeship for FAC22. @Albadylic to lead on working out the details with @oliverjam.

Albadylic:
Now that we have mapped the 3 weeks to the curriculum, @Albadylic 's next action is to map meeting PO's, Discovery and User Research to the first 6(?) weeks in preparation for TfB projects. 

Relates to #468

sofer:
Going to General Circle.

Albadylic:
https://www.notion.so/foundersandcoders/A-new-model-for-Tech-for-Better-fe9f5e8fd86f4205b21b424611703a66

Issue #452, opened 2021-05-07
Scheduling time for testing and estimation


sofer:
The project teams are struggling with finding time to do estimation and testing. Perhaps we ought to schedule 30 mins on Wednesday at 2pm to do story estimation and 30 mins on Thursday at 10am to set up tests. 

Scheduling time may be the only way that we can ensure some attention is paid to these two hugely important tasks (arguably the two most important skills to be learnt at FAC) and making the time very short means that it won't take up an inordinate amount of project time. Just getting into the habit of doing these things at about the right time in the project, even if it means only very rough estimating and writing a couple of tests, is going to be worth it.

sofer:
Make Wednesday more structured:
- decide on scope
- set up repo
- write user stories
- do software estimation
- set up tests
- set up production environment

Trial it for next week and then use it for FAC22

oliverjam:
@oliverjam  to put all this stuff into the schedule for Build Week 1 Monday

PR #510, opened 2021-09-03
Fac22 schedules

Albadylic:
So this is where I got to with rejigging the schedules for the first few weeks. There's a couple of things to consider:
- Adding the web science curriculum (#475)
- Adding Tech for Better (#473)
- Adding time for testing and estimation (#452)
- Adding some time for inductions in the first week
- There's a bank holiday schedule for Server Side App which won't be a BH for FAC22

### DIFF
--- a/src/_data/defaultSchedule.yml
+++ b/src/_data/defaultSchedule.yml
@@ -39,12 +39,16 @@ tuesday:
   - start: 13:00
     end: 14:00
     name: Lunch
-  - start: 16:00
-    end: 17:00
-    name: Interview practice
-  - start: 17:00
+  - name: Project intro
+    start: 17:00
+    end: 17:15
+    type: project
+    url: ../project
+  - name: Project set up, estimation and testing
+    start: 17:15
     end: 17:45
-    name: Speaker
+    type: project
+    url: ../project
   - start: 17:45
     end: 18:00
     name: Check-out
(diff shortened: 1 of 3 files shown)


src/course/syllabus/server/schedule.md
```diff
@@ -22,31 +22,26 @@ schedule:
...
-      start: 11:00
-      end: 12:30
     - name: Server-side forms
-      start: 12:30
+      start: 11:00
       end: 13:00
       url: /workshops/server-side-forms/
       type: workshop
-    - name: Server-side forms
+    - name: Tech for Better - Introduction
```
oliverjam, line 30: The design for the schedule means long names don't fit particularly well. Could we abbreviate these to `TFB -`?

src/_data/defaultSchedule.yml
```diff
@@ -39,20 +39,28 @@ tuesday:
...
-  - start: 16:00
-    end: 17:00
-    name: Interview practice
-  - start: 17:00
+  - name: Project intro
+    start: 17:00
+    end: 17:15
+    type: project
+    url: ../project
+  - name: Project set up, estimation and testing
```
oliverjam, line 47: I'm not convinced this can all fit into half an hour. I remember recent cohorts easily spending half a day getting every ready (especially once they started using databases). Maybe this should just be "Repo setup & estimation", to imply they should be focusing on GH issues rather than code?

PR #518, opened 2021-09-29
Code of Conduct should link to main FAC site

oliverjam:
So we don't maintain two copies and let one get out of sync


Albadylic (approved):
Sorry, only just caught this. Thanks - looks great!

PR #525, opened 2021-10-13
FAC23 server-side app week

oliverjam:
- Update the schedule to remove the bank holiday
- Clarify the project criteria and move spikes to separate page

Screenshot 2021-10-13 at 13 10 35

### DIFF
--- a/src/course/syllabus/apprenticeship/server-side-app/spikes.md
+++ b/src/course/syllabus/apprenticeship/server-side-app/spikes.md
@@ -1,3 +1,5 @@
+Try to create a working MVP on glitch.com or stackblitz.com. This will be helpful to show during your presentation and to reference in your projects.
+
 ## File uploads
 
 Allow users to upload files to your server (like images or audio clips).  


src/course/syllabus/apprenticeship/server-side-app/schedule.md
```diff
@@ -2,22 +2,30 @@
...
       url: /workshops/express-middleware/
       type: workshop
+    - name: Project intro
+      start: 14:00
+      end: 14:15
+      url: ../project
+    - name: Spike research
+      start: 14:15
+      end: 17:45
+      url: ../spikes
```
Albadylic, line 17: Would we like to break this time up more?

It feels like 3.5 hours could be a long time for them to be working without a structure suggested for how to spend their time. 

We could break the time into something like: decide a topic, research, get an MVP solution working. We could say this at the top of the spikes file. 

Since they have extended project time maybe it's okay not to give too much direction and leave it up to them to decide how they'd like to structure their time. It might be enough to let the mentors know to check in often and encourage them not to get stuck deciding or researching for too long.

Albadylic (approved):
Looks like a positive change!

PR #560, opened 2022-01-13
change unhandledRejection to camelcase

reubengt:
having unhandledRejection in camelcase in the example and then all lowercase in the explainer text below could be confusing, node expects it to be camelcase right?


oliverjam (approved):
Good catch, thanks!

PR #566, opened 2022-01-17
update tech spikes to 3 a week for FAC23

reubengt:
following discussion w @oliverjam : keep the tech spikes most relevant to project work.
The following spikes are excluded.
Server : Modules  (already covered in intro talk)
Database: SQL vs noSQL (already covered in intro talk)
Auth: HTTPS/TLS (likely will never implement TLS stuff as a junior dev: hosting providers do it for you)


oliverjam (approved):
Looks great, thanks ✨

PR #574, opened 2022-01-25
Update employment prep schedule

### LINKED ISSUE #573: The employment prep session schedule need updating
We originally put something there that roughly approximated what I did last term, but what I am delivering this term looks different.

sofer:
Fixes #573


sofer:
I might change these around a bit, but the default schedule is the same for each week now after the first couple of weeks.

oliverjam (approved):

PR #583, opened 2022-02-09
move role circles to thursday mornings

reubengt:
The cohort have said that they much prefer role circles to be part of the standup on the second day, once they've had a chance to work on their projects a little.

Database week is a weird one because of the web science talk, and for the two weeks  with three day projects I've kept it on wednesday as that is day 2. (next and server side week)

### DIFF
--- a/src/_data/defaultSchedule.yml
+++ b/src/_data/defaultSchedule.yml
@@ -73,7 +73,11 @@ thursday:
   - start: 9:45
     end: 10:00
     name: Check-in
-  - start: 10:00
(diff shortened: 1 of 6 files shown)


oliverjam (approved):

Albadylic:
This looks good - my only suggestion would be that if we want them in 5 weeks and not in 1 week it might be better to put this in default schedule and remove from the 1 week.

reubengt:
oops yeah that makes sense! I'll change the default schedule instead

Issue #590, opened 2022-02-25
Pre-apprenticeship: last three weeks.


sofer:
It might be a good idea to pin something down ahead of time other than just interview preparation. We have talked about getting them to do side-projects. I also did an exercise in the classroom yesterday that we may be able to build on about developing habits of study. The responses to that exercise are in the `spr-22-classroom` channel. We could perhaps split them into groups, let them decide on different topics (any topic!), get them to research different learning resources, agree on a programme of study for themselves and then finish off with some sort of in-person group project over a whole day (or two).

Albadylic:
We need constraints around what people will work on in the 5 week break - suggesting front-end libraries they could explore, and projects they might like to work on. We'll provide a shortlist of libraries to look at.

They can also focus on revisiting the learning outcomes for the weeks previous.

Workshops for the last three weeks could focus on planning, setting up and building some (or all) of these projects. Building a habit of code-reviewing each other's projects and how to pick up and work on a project solo. We might like to follow a similar model of the existing curriculum: week 1 is to learn the library, week 2 is to plan the project, week 3 is present the project & code review.

sofer:
Some libraries that might be worth exploring:
- https://d3js.org/
- https://threejs.org/
- https://aframe.io/
- https://www.chartjs.org/
- https://animejs.com/
- https://leafletjs.com/
- https://pixijs.com/
- https://phaser.io/
- https://www.babylonjs.com/
- https://ar-js-org.github.io/AR.js-Docs/

sofer:
Also, I am aiming to schedule Q&As 5.10-6pm in all 6 (2 so far) sessions in the final 3 weeks.

sofer:
Just confirming: we are going to be very busy with screening questions and take-home challenges in the last three weeks, so it makes sense for the final "side-project" to be low intensity. Larger teams may also work better, so that the teams don't stall due to all team members prioritising interview prep.

PR #598, opened 2022-03-11
Add Projects topic to SB3 schedule

oliverjam:
Added a selection of 4 projects so far, plus the schedule for next week. We'll need to update with the following weeks once we know exactly when Q&As are etc.

My plan is for them to work through the revision quiz during Monday's session, so they can figure out what they need to focus on and pick a project accordingly. Then they can work on the project in their free time, and we can do code review/help sessions on Mondays.

Relates #590


sofer (approved):
Very nice.

PR #608, opened 2022-03-31
Simplify database setup in pg workshop

### LINKED ISSUE #570: .env files and the dotenv node module - cover this explicitly in a workshop?
FAC22 and FAC23 have run into the same issue of not knowing how to import environment variables. 

https://learn.foundersandcoders.com/workshops/node-postgres/

This workshop creates a .env file with a script but doesn't explain that the dotenv module is what actually reads that file.

I don't think it needs to be a separate workshop, can just be an added line of explanation into what happens with the .env file?

oliverjam:
I realised the database creation/population scripts were just confusing extra overhead for people. They do very little since you can condense them down into basically just `createuser x && createdb x --owner x`.

Also explain .env properly to resolve #570. I'm not happy with how verbose this ends up being, but unfortunately unless we have an intro to env vars prior to this we need to explain quite a few things for `dotenv` to make sense.

I think a general principle I want to try to follow moving forward is "don't install new stuff for them". If we're introducing a new library for the first time they should type the `npm i` themselves to ingrain it, and so they don't forget to do that when they start their project 😅


src/workshops/node-postgres/index.md
```diff
@@ -76,27 +76,47 @@ This script executes the SQL commands inside `/database/init.sql` against your D
...
+```
 
-Take a look at `/database/connection.js`. We import the `pg` library, then connect to the right DB by passing in a "connection string" (a URL identifying the DB and user).
+You _could_ hard-code this URL into our app code, but this address is only correct for the database running on your computer. When someone else clones your repo they'll have their own DB set up.
 
-{% disclosure "More info about the connection string" %}
+### Environment variables
 
-The connection string is a URL pointing to your exact Postgres DB, including the user and password. It is structured like this:
+It's best to read configuration like this from an "environment variable" (env vars). This is a value set in the shell before you start your application. For example it's common to read the `PORT` env var to know what port your app should listen on.
```
reubengt, line 95: "This is a value that is made available to the terminal when you start your application"?

"set in the shell" doesn't feel explicit enough😅

src/workshops/node-postgres/index.md
```diff
@@ -76,27 +76,47 @@ This script executes the SQL commands inside `/database/init.sql` against your D
...
-When we run `npm run dev` to start the server it will execute `nodemon -r dotenv/config server.js`. This tells Node to use the [`dotenv`](https://www.npmjs.com/package/dotenv) library to load env vars _before_ our server starts. This will read the `.env` file and load all the env vars automatically.
+Then change your `dev` npm script in the `package.json` file to this:
 
-{% enddisclosure %}
+```
+"nodemon -r dotenv/config server.js"
+```
 
-The `/database/connection.js` file exports a "pool" of connections for our app to use. This helps handle multiple requests at the same time. We can import the `db` variable in any other file to send requests to the DB.
+The `-r dotenv/config` bit tells the `dotenv` library to read the `.env` file and pass all the values inside it to your application. They will then be accessible via `process.env.VAR_NAME`.
```
reubengt, line 119: "you can access them in your code via `process.env.VAR_NAME`"

just to be more explicit since there is quite a bit of jumping between terminal, package.json, and application code here.

reubengt (changes requested):

src/workshops/node-postgres/index.md
reubengt: Looks good otherwise! Definitely think "don't install for them" is a good idea😅

src/workshops/node-postgres/index.md
oliverjam: Yeah there is no accurate way to describe this without the context of how programs actually run :/ I think "made available to the terminal" might be a bit misleading, since the terminal application isn't really involved.

Maybe an example would make it clearer?

> This is a value you can set in your shell before you start your application. E.g. `PORT=3000 node server.js`. You server can then read this value to know what port your app should listen on.

reubengt: yep I like the example! also adds some more clarity to the need for dotenv over doing it manually

PR #622, opened 2022-05-16
Initial WIP schedule changes for SB4

oliverjam:


sofer:
Also, be aware that from Week 4 there will an employer Q&A on Thursdays, which will cut into workshop time.

PR #624, opened 2022-05-16
increase tfb to 90 minutes

reubengt:
took 15 mins out of project setup and 15 mins out of the workshop right after tfb (in some cases 30 mins out of build time)


Albadylic:
Thanks for pushing this. I think we might have said we'd move project set up to being first thing on Wednesday. I think we'd prefer that over removing workshop time, and where we have an additional 15 minutes left over we could extend the workshop by that?

reubengt:
don't think there's any point specifically marking project setup time if its wednesday morning, have just set it up to say "project"

Albadylic (approved):
Amazing, thanks!!

PR #633, opened 2022-05-19
Update schedule for bank holiday

reubengt:
Have shortened the express middleware workshop from 3 hours to 2 - would this be enough?


oliverjam (approved):
I'll update the middleware workshop to be auth only soon

Issue #532, opened 2021-11-09
Introducing the terminal


Albadylic:
It feels like introducing the terminal when we introduce Git and GitHub as part of the application prompts is too big a jump in concepts. We might like to introduce the terminal more formally as part of the prompts.

oliverjam:
Yeah too many concepts at once. I did introduce one cohort of applicants to Git via just VS Code's GUI ages ago. There are a few difficulties with that though, cause it can't quite do everything yet

Albadylic:
Introducing the terminal relies on folks having an idea of how their computer works - e.g. file structure.

Minimum solution is probably to add the command-line-primer or something similar to the application prompts. We should intro it one way or another.

We might like our own resource covering the fundamentals when we have time to create it. 

The connection between local / remote is more important than understanding file structure.

Albadylic:
Relates https://github.com/foundersandcoders/tasks/issues/73

oliverjam:
Do we want to shift the [Digital Ocean command-line primer](https://www.digitalocean.com/community/tutorials/a-linux-command-line-primer) forward from the Skills Bootcamp pre-course?

Albadylic:
fixed by #643

PR #643, opened 2022-05-24
Git

Albadylic:
This directly relates to the Tribute Page project as an introduction to Git. We could also add the terminal intro, and Git via terminal workshops to Project Four. But that's not as high a priority as having _something_ introducing Git and GitHub in a lightweight sense for the first project.


Albadylic:
Add staging and committing and pushing definitions

src/workshops/applicant-git-1/index.md
```diff
@@ -0,0 +1,87 @@
...
+
+## Local vs. Remote
+
+We'll describe repositories in terms of local and remote versions.
+
+A _local_ repository is a folder on your computer which you save files to. A _remote_ repository is a folder online which your files can be duplicated to.
+
+## Git Flow
+
+_Git flow_ refers to the process of synchronising code between a local and remote repository.
```
oliverjam, line 35: I think we should use a different term to describe this to be precise. Git flow/gitflow refers to quite a [specific methodology](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow) that we definitely don't teach/follow/encourage.

We roughly teach Git _Hub_ Flow (https://docs.github.com/en/get-started/quickstart/github-flow), but it seems like this section is more generally referring to just a general "workflow" for working with Git/GitHub.

src/workshops/applicant-git-1/index.md
```diff
@@ -0,0 +1,87 @@
...
+
+![VS Code offers the user to publish to a private or public repository on GitHub](./images/9.png)
+
+And that's all! You'll now have published the files to GitHub. If you visit your profile on [GitHub](https://www.github.com) you should now see your repository.
+
+![A repository on GitHub](./images/10.png)
+
+From here, you can continue to stage, commit and push your changes.
+
+There's no hard and fast rule for how often you should commit. This might be after every 20 lines of code; every 30 or 60 minutes; or after each feature you complete. Getting into the habit of making small commits often will give you a good level of practice with _Git flow_ as well as GitHub activity on your profile (green squares). Regularly pushing your changes will ensure your codebase is backed up and version-controlled.
```
oliverjam, line 87: I think it might be more helpful to frame this like "you should commit once you have completed a change or feature, not after a certain amount of code". Saying "there's no rule" is possibly misleading, especially when they are total beginners and really want explicit guidance.

src/workshops/applicant-git-2/index.md
```diff
@@ -0,0 +1,82 @@
...
+When staging, you can specify files to stage, or using the `.` you can stage all the files in your current directory.
+
+```
+git add index.html
+git add .
+```
+
+Commiting is like saving your progress or taking a snapshot of the current codebase. You are telling Git that all the changes made should be tracked. The repository will be saved at the point in time you commit.
+
+Running the command without the `-m` flag will open up an editor in your terminal where you can write a commit message. Exit this by hitting `esc` and typing `:wq`.
```
oliverjam, line 38: Is it worth mentioning here that they can [change their default terminal editor](https://oliverjam.es/blog/make-your-terminal-nicer/#set-your-default-editor) to avoid Vim opening every time they commit?

It's quite nice just typing `git commit` and having VS Code open

oliverjam (commented):
This is looking amazingly detailed, thank you ✨ 

I am a little concerned about adding a bunch of images to the repo, for a couple of reasons

1. It's hard to maintain image content as apps/UIs change and options move around etc. They will inevitably be out of date soon
1. It's hard to make images consistent, especially across different contributors with different systems/wallpapers etc 😅 
1. Storing binary files like images in Git is ...not ideal. The repo is currently about 20MB—adding these 9 images will ~double that to 45MB.

My suggestions would be:

1. All screenshots should be "full window", with no other tabs or background stuff to distract. You can do that on a Mac by hitting `cmd-shift-4` to enter "screenshot mode" (your cursor changes to a targeting reticule), then hitting `spacebar`. This should let you click a specific window to screenshot _just_ that without any background or other desktop stuff.
1. Resize all images to a reasonable resolution (i.e. if you screenshot the whole desktop by default you get a giant PNG that is thousands of pixels wide, whereas these will only ever be displayed in an ~800px wide column on the site). You can resize images in Preview quite easily. It's probably worth saving as a jpeg too since that should be way smaller than png.
1. Run images through [ImageOptim](https://imageoptim.com/mac) to reduce the file size _a lot_

Albadylic:
This is super helpful, thanks @oliverjam . Happy to take your suggestions and make changes to this tomorrow.

I didn't know you could use `spacebar` to get just the window, I've only used `cmd-shift-4` to take screenshots of inconsistent sizes 😅 🙃

oliverjam:
I've just discovered that there's a full GUI Screenshot app now. Hit `cmd-space` then search for "Screenshot"—it gives you a little bar overlay with all the options for different types of screenshots exposed (and even screen recording).

Albadylic:
Introducing terminal #532 :

```
{% box %}
For this project, we also expect you to research how to use git from the command line, instead of using the source control tab on VS code.

If you have not used the terminal/command line before - it's also a good idea to do some research on navigating your file system using the terminal, and using commands like `cd`, `ls`and `touch`/`echo`
{% endbox %}
```

oliverjam (approved):

PR #660, opened 2022-06-08
fix tests failing at start

reubengt:
Joe found that tests weren't passing in this workshop because of a syntax error (a getPokemon function was imported when already declared lower down in the file)

Also updated the test command to jest --watchAll because --watch only works for git repos

### DIFF
--- a/src/workshops/react-refactor-classes/starter-files/workshop/Pokemon.jsx
+++ b/src/workshops/react-refactor-classes/starter-files/workshop/Pokemon.jsx
@@ -9,6 +9,7 @@ function getPokemon(name) {
     })
     .then((res) => res.json());
 }
+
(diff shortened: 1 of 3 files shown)


src/workshops/react-refactor-classes/starter-files/workshop/Pokemon.jsx
```diff
@@ -1,6 +1,14 @@
 import React from "react";
-import { getPokemon } from "./utils";
```
reubengt, line 2: this was causing tests to fail

oliverjam: Weird how this hasn't been an issue for prior cohorts. Don't think I've touched this workshops for ages

Issue #612, opened 2022-04-06
Split out projects and github workflow pages of handbook


sofer:
Most teams do find good ways of working, but a page of hints on how to work well together as a team, such as switching up pairs regularly, reminding each other to take breaks and the importance of the whole team coming together for reviewing and merging PRs, might be a useful addition to the coursebook.

reubengt:
https://learn.foundersandcoders.com/course/handbook/projects/ I made an update last term so this is briefly covered [here](https://learn.foundersandcoders.com/course/handbook/projects/)
but we could separate this into another longer document.

sofer:
@reubengt suggests two pages and will aim to do a pr before the next circle.

PR #661, opened 2022-06-08
separate github workflow to a second document in coursebook

reubengt:
closes https://github.com/foundersandcoders/coursebook/issues/612


src/course/handbook/github-workflow.md
```diff
@@ -0,0 +1,41 @@
+### GitHub workflow
```
oliverjam, line 1: This needs to be an h1 now it's a full page title

```suggestion
# GitHub Workflow
```

Issue #657, opened 2022-05-31
Timetable clash in Week 8


sofer:
The employer day is on Tuesday "morning (and lunch)" and is likely to run into the Tech for Better workshop at 2pm. One employer has already requested the afternoon instead. We may want to move Tech for Better to a different day this term.

Also, in the light of #653, we may want to move the employer day to a Monday from next term.

Albadylic:
Reach out to PO's by end of the week

PR #664, opened 2022-06-08
Proposed schedule for last day of FAC24

### LINKED ISSUE #610: Last week of Tech for Better
The showcase for Tech for Better falls outside of the programme. This presents a few problems and makes it more difficult to coordinate than it needs to be.

I think we should move the schedule for the second build week around to give time for the dev's to prep their presentations earlier, and to add the thoughts of their PO during the sprint. We'd need to move some of the stuff from Friday morning into Thursday evening and have a hard deadline for the project build to finish.

Albadylic:
Relates https://github.com/foundersandcoders/tasks/issues/209

Fixes #610 

- [x] Remove speaker from last day - replace with 'Graduation party'

Whether or not we get Newspeak, I think I'd propose we go with this schedule for FAC24

I also snuck fixing #657


reubengt:
really not in favour of reducing build time further than 4 days. can we remove "presentation practice" and treat the test run as practice? 

that way we can reclaim some more of thursday evening.

reubengt (changes requested):
really not in favour of reducing build time further than 4 days. can we remove "presentation practice" and treat the test run as practice?

that way we can reclaim some more of thursday afternoon

Albadylic:
For Thursday afternoon, we'd have the Product Owners around the prepare with them. We can then give notes on the presentation before they deliver it for a wider audience on Friday. I think people would be grateful for a chance to practise before delivering it for the first time. 

I forgot we needed to remove the employment morning from the week before which regains them 0.5 of a day.

reubengt (approved):

reubengt:
happy with this 😄

PR #673, opened 2022-06-14
Add flag to use npm instead of yarn

reubengt:
Yarn is the default installer for create-next-app

FAC24 ran into a yarn.lock file after creating their next project - was confusing as they hadn't encountered yarn before.

I've updated this to pass in a flag to explicitly install extra dependencies for create-next-app with npm instead of yarn.


oliverjam:
Oh interesting. Surely it could only use Yarn if they had Yarn installed on their machine?

reubengt:
yeah i'll ask who initiated the project. probably only defaults to yarn if its installed.

oliverjam:
Probs still best to ensure it always uses npm though yeah

Issue #630, opened 2022-05-18
Fundamentally restructure the timetable


oliverjam:
WIP issue, will add more info later

We've had the same structure forever, so maybe it's time for a change. How about making Monday a work-from-home-do-the-workshops-yourself day, and Tuesday a discuss-workshops-all-together day?

sofer:
This new approach would meet three objectives that we do not currently do very well:
- Allow learners to spend time on self-study
- Allow them more opportunity to master the material at their own speed
- Allow us to flip the classroom, i.e. come together to discuss material already studied and to provide remedial support.

It could also potentially reduce the number of required in-class days to one! 

I would be in favour of doing a trial week this term using some version of this approach, perhaps setting all the workshops on Friday afternoon and leaving them to it until Tuesday.

sofer:
If we did it this term, the obvious week to go for would be Week 6. It's a bit of a cheat, because it's not quite like earlier weeks, but there is more-or-less nothing going on that week on Monday other than workshops, so we could just leave them to it.

We could also give the Week 1 workshops as homework over the holidays to FAC25, which would give us more time on Monday to do the Welcome talk and then to crack on with workshop reviews. If it goes well, we could adapt the approach to the following weeks, too.

sofer:
Some feedback from a FAC24 learner:
- In general, having individual study time would suit them.
- Is one day enough for all workshops? Perhaps give over some time on Tuesday morning for finishing them off or doing the final one.
- Consider having Tuesday tutorial groups based on the project teams, so that they can support each other even before the projects kick off.
- It might be worth doing the project and spikes intro last thing on Friday, so that learners are thinking about the project and have picked their roles before they start on the workshops.

Albadylic:
I was concerned the workshops might not fit into one day, but looking in detail at the schedules for weeks 1 -6, I'm now convinced they can.

| Week | Workshops | Spikes | Project time | Morning challenge | Total | 
| --- | --- | --- | --- | --- | --- |
| 1 | 8.5 | 1 | 11.5 | 1 | 22 |
|2 | 8.25 | 1 | 11 | 1 | 21.25 |
| 3 | 8.5 | 1 | 10 | 1 | 20.5 |
| 4 | 2.83...| 3.5 | 17 | 0 | 23.33...|
| 5 | 11.5 | 0 | 11 | 1 | 23.5 |
| 6| 6.25 | 0 | 17.25 | 0 | 23.5 |

_Times are in hours_

[...]

PR #676, opened 2022-06-15
Add new default schedule - to merge after FAC24 end of term

Albadylic:
I'm just opening this so we can view the schedule in CC when we chat about #630 - this doesn't want to be merged yet.


src/course/syllabus/apprenticeship/client-side-app/schedule.md
```diff
@@ -2,52 +2,9 @@
...
-      url: /workshops/react-refactor-classes/
-      type: challenge
-    - name: Project
-      start: 11:00
-      end: 13:00
-      type: project
+    - name: Choose a project idea
+      start: 15:45
+      end: 16:00
+      url: https://github.com/fac25/final-project-ideas/issues
```
Albadylic, line 9:
```suggestion
```

This will need to be updated for every cohort. It may be easier not to link to a repo

src/_data/defaultSchedule.yml
```diff
@@ -1,61 +1,52 @@
...
     name: Lunch
-  - name: Project setup
-    start: 17:00
-    end: 17:45
+  - name: Tech for Better
+    start: 14:00
+    end: 15:30
+  - name: Final project prep
+    start: 15:45
+    end: 16:00
```
Albadylic, line 33: There's a 15 minute gap between TFB ending and the project-mgmt stuff starting. Is this deliberate?

src/_data/defaultSchedule.yml
```diff
@@ -1,61 +1,52 @@
...
+    end: 10:10
+    name: Feedback survey
+  - start: 10:10
     end: 11:00
-    name: Spike presentations
-    type: presentation
+    name: Morning challenge
+    type: challenge
+  - start: 11:00
+    end: 12:30
```
Albadylic, line 18:
```suggestion
    end: 12:20
```

src/_data/defaultSchedule.yml
```diff
@@ -1,61 +1,52 @@
...
+    name: Group discussion
+    type: workshop
+  - start: 12:30
```
Albadylic, line 21:
```suggestion
  - start: 12:20
```

Albadylic: 80 - 40 time split

Albadylic (commented):
Thanks for putting this together. I left a couple of suggestions.

We need to add the weekly intro somewhere on Friday, or include a video as a resource. We might want to add it after SGC and before the speaker.

src/course/syllabus/apprenticeship/client-side-app/schedule.md
reubengt: cool cool

src/_data/defaultSchedule.yml
reubengt: a break after tech for better is nice to have😅

src/_data/defaultSchedule.yml
```diff
@@ -1,61 +1,56 @@
...
   - start: 9:45
     end: 10:00
     name: Check-in
   - start: 10:00
-    end: 10:15
-    name: Spike prep
-    url: /course/handbook/spikes/#what-makes-a-good-research-presentation
-  - start: 10:15
+    end: 10:10
+    name: Feedback survey
```
oliverjam, line 12: It feels a little weird to have this here. Could we not just ask them to do it during their Monday on their own? Or would we end up with people skipping it do you think?

src/_data/defaultSchedule.yml
```diff
@@ -1,61 +1,56 @@
...
+  - start: 10:10
     end: 11:00
```
oliverjam, line 14: I also think people struggle enough with morning challenges for a full hour. Cutting 10 mins out will make them even worse 😅

oliverjam (commented):
This is great—love how much simpler the schedules for most weeks get. I left a couple of questions around Tuesday morning, which might be a little packed atm

src/_data/defaultSchedule.yml
Albadylic: I think we put it in the schedule only to make sure it gets completed. It could easily be part of the Friday or Monday activities. Many cohorts have asked for it to come before the weekend. But I would be surprised if this didn't add a bit of chasing to the CF's workload.

Albadylic:
> This is great—love how much simpler the schedules for most weeks get. I left a couple of questions around Tuesday morning, which might be a little packed atm

I've moved the feedback to Friday just before the speaker. I think this will be a good wee spot, to give a break between the two talks, and while things are still fresh. This gives back 10 mins to MC and takes 10 mins out of SGC (which should be fine within 50 mins)

Albadylic:
I do wonder if there's another way to render the specific weekly stuff, rather than having a schedule for each week.

There's three things that change:
- Tech for Better activity (not currently displayed what happens week to week)
- What the intro talk will be (i.e. next week's topic)
- What 'final project prep' is to be done (e.g. draft a project idea)

I wonder if we could implement an ordered list for each of these three and then render the specific ones week to week. It may be simpler than updating the schedules for each week to include these specific activities. Although, it's probably not a priority over merging this - which works for us just now.

oliverjam:
Yeah having a whole separate schedule to merge into the default made sense when each week was more different than the same. Now all we really need to define for each week are:

1. Link to workshop
2. Link to morning challenge
3. TFB activity
4. Intro talk
5. Final project prep

Assuming those are always at the same time each week (?) then yeah we can just define them simply instead of as part of a schedule.

If we merge this I can simplify that at some point later.

oliverjam (approved):

PR #681, opened 2022-06-22
Give stronger hint for server form filter solution

### LINKED ISSUE #620: Server week workshop feedback
- Server-side forms workshop - people got stuck on the "search" challenge. The focus on learning how to handle url search params was overlapped a little by people trying to understand the object/string manipulation involved in that search task (although I do think it's at an appropriate level for the full time program). A suggestion from Joe was to move that challenge further down and make sure people get through GET and POST handling first. Also I think we can extend this workshop for another after lunch - the TFB introduction can probably be shortened?

- Cypress - install in pre-requisites, people said the workshop itself could have been shortened a little (put the extra time in the first workshop)

@oliverjam thoughts? I think the suggestion to move the search challenge to later in the workshop to prioritise get and post makes sense.

oliverjam:
@reubengt do you think this is enough? I feel like the only way to provide any more of the solution would just be giving the whole answer

Fix #620

### DIFF
--- a/src/workshops/server-side-forms/index.md
+++ b/src/workshops/server-side-forms/index.md
@@ -130,8 +130,11 @@ You can use `string.includes` to check if a string contains a given substring. E
 
 ```js
 const search = "rov";
-const dog = "rover"
-if (dog.includes(search)) ...
+const name = "rover";
+const match = name.toLowerCase().includes(search.toLowerCase());
+if (match) {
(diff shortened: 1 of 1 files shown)


reubengt:
yeah this looks fine, we'll see how it goes for the next cohort. sorry it took so long to look at this pr🤦‍♂️

reubengt (approved):

PR #718, opened 2022-08-30
Application review

Albadylic:
- Add a new Pet Shop project, replace Fruit Shop
- Update some of the guidance on a couple of projects
- Add a guide to getting started


src/course/syllabus/application/pet-shop/project.md
```diff
@@ -0,0 +1,58 @@
+For this project, we're giving you some data in a JavaScript object. Create an interface where the user can interact with and change what is displayed on your webpage. Interacting with page will alter and update the data stored in the object.
```
oliverjam, line 1:
```suggestion
For this project, we're giving you some data in a JavaScript object. Create an interface where the user can interact with and change what is displayed on your webpage. Interacting with the page will alter and update the data stored in the object.
```

src/course/syllabus/application/pet-shop/project.md
```diff
@@ -0,0 +1,58 @@
...
+
+- [ ] Store the object below in a JavaScript file in your codebase
+- [ ] Render the data into HTML using DOM Manipulation
+- [ ] Allow the user to update object values when interacting with the webpage
+
+## Data
+
+Your data is stored in the following object:
+
+```
```
oliverjam, line 15:
```suggestion
```js
```

src/course/syllabus/application/pet-shop/project.md
```diff
@@ -0,0 +1,58 @@
...
+let dataObject = {
+  1: {
```
oliverjam, line 17: I'm not sure an object with number keys is a great idea for a beginner thing. This is effectively recreating an array, but without all the convenient methods arrays normally have.

If each dog has a unique name I would recommend using that as the key instead, as it's a more realistic example of using objects. E.g.

```js
let dataObject = {
  fido: {
    name: "fido",
    type: "dog",
  }
}
```
Up to you whether to repeat the name again in the object—it can make using the data easier, but it is duplication.

Albadylic: Yeah, I'm not sure here what the best approach is. The most natural feels like it would be to include an array of objects. Or, as you say, using unique keys for each pet. Do you have an opinion on which is better practice?

Albadylic: Perhaps a simpler object is a better approach, for example:

```js
let user = {
  name: "Gregor",
  github: "Albadylic",
  age: 28,
  hobbies: ["pickling", "fermenting", "making kombucha", "cooking", "playing Overcooked"]
}

Albadylic:
Based on Oli's feedback, I've made a change to the new project.

Reflecting on it now, I could have kept the pet's name as the key, but hey I went with Wes Anderson films instead.

We've got an object, with nested objects for each film. The project is to build a user interface and display all the data. The applicant can do this however they like. Hopefully this project will give us some more interesting implementations than the Fruit Shop!

oliverjam (approved):
I like the movies ✨

PR #719, opened 2022-09-07
switch employer day in schedule from week 3 to week 4.

ByteSizedIT:


ByteSizedIT:
Change confirmed with Gregor

Issue #716, opened 2022-08-03
Combine workshops into single resource for each topic


oliverjam:
Now that we're asking learners to complete all their learning on the first day of the week solo it makes more sense to have a single mega-resource, rather than shorter separate workshops.

This will save a bunch of "setup" and "cooldown" time wastage at the start/end of smaller workshops, and allow learners to build up a more interesting thing by the end.

This shouldn't require a huge amount of work beyond concatenating markdown files, but we will have to consider how to make the actual workshop practice/starter code work.

I think we should structure it as two separate halves:
1. "Learn"—a longer-form document explaining the required concepts (maybe with an accompanying video #650)
2. "Practice"—a repo with starter code and a series of challenges (and tests to verify #690)

PR #720, opened 2022-09-08
Update schedule for FAC25 Server Week

### LINKED ISSUE #572: Replace debugging challenge with server validation
This is a very big, very important missing piece from our Node weeks. We never really touch on validating or sanitizing user-submitted data. It has come up in my live code reviews for multiple cohorts, so we should really have a workshop on it. Examples:

1. Relying on client-side validation for sign up. I demonstrated POSTing a request with `curl` that had no email address. This broke quite a lot of things in their app since the code assumed a user would always have an email.
2. Dumping user input straight back onto the page. I demonstrated submitted a string with a `<script>` tag that could do malicious things.

I don't think this has to be super in-depth—a morning challenge using a simple library like [Superstruct](https://docs.superstructjs.org/) would be quite fun.

### LINKED ISSUE #618: Replace Cypress with simple Node testing
Should we be considering an alternative?

oliverjam:
Combines Express/testing learnings into one workshop + challenge (relates #716)

Replace Cypress with vanilla Node testing
(fix #618)

Replace debugging challenge with sanitize/validate one (fix #572)

### DIFF
--- a/src/course/syllabus/apprenticeship/server/schedule.md
+++ b/src/course/syllabus/apprenticeship/server/schedule.md
@@ -2,7 +2,17 @@
 layout: schedule
 schedule:
   monday:
+    - name: Server workshop
+      url: /learn/server
+      type: workshop
+      start: 9:45
+      end: 18:00
   tuesday:
+    - name: Sanitize & validate challenge
+      url: https://github.com/foundersandcoders/sanitize-validate-challenge
+      type: challenge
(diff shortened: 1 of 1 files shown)


Albadylic (approved):

Albadylic:
For this week, the challenge is contained within the workshop. Are you still thinking about using GitHub classroom for future weeks?

PR #721, opened 2022-09-09
Reduce spikes to 1 per project

### LINKED ISSUE #691: Reformulation of tech spike and project structure
See https://github.com/foundersandcoders/coursebook/issues/630#issuecomment-1170032078:
> ## Projects
> 
> - Tech spikes are not relevant enough to the projects right now
> - Projects should have at least one (or a few options of) technical problems to solve beyond the week's "boilerplate"
> - We might want to consider much more specific project ideas so people don't get lost at the creative step
> - A better project structure might look like:
>    1. Solve the technical problem (a spike with throwaway code)
>    2. Come up with an idea/plan the project
>    3. Build the actual features (should be simpler since the hard part is already solved)
>    4. Respond to feedback (after code review)
>    This is more similar to the structure of a real project, where you wouldn't just dive into building something from the start

oliverjam:
Move spike to project intro to make it clear they happen together. Basically I'm trying to frame it as "this is something you have to figure out before you start".

Also move some old spike resources to resources tab so they aren't lost.

fix #691

Screenshot 2022-09-09 at 15 18 50


src/course/syllabus/apprenticeship/database/project.md
```diff
@@ -1,5 +1,19 @@
...
+## Spike
+
+Before you start writing features you need to design the schema for your data. Think about what different things your app needs to store, how they relate to each other, and how you can avoid duplicating information. Record your schema in your `README.md` using Markdown tables.
+
+### Questions to consider
+
+1. What kinds of data relationships are there?
+1. What's a foreign key? How can they help us design schemas with relational data?
+
+### Useful resources
```
Albadylic, line 12: We might like to include a tool for creating schemas (e.g. https://dbdiagram.io/home). Although, this may be less relevant if we're encouraging them to include their schema in their README.

Albadylic (approved):

src/course/syllabus/apprenticeship/database/project.md
oliverjam: Yeah I did consider that, wanted to keep it simple. I guess those diagrams help people visualise relations—no reason they can't embed an image in the README

Albadylic:
Sounds good! 👌 

For 2. - I think the only thing we need the documentation for is the first time we ask a cohort to do spikes, and a CF to introduce them. It being concise is definitely a good thing.

ByteSizedIT (approved):
Hey Oli, 

Like Gregor has said, the change in format looks good 🤩, much more streamlined. 

Nothing more to add other than highlight there is a few outstanding references to Postgres in the Project/Spike intro in some of the later weeks.

Without wanting to labour the subject - is Heroku used for the final TFB projects very often?

Thanks for all your work on this and the workshops 👍

oliverjam:
Oh good catch, I'll pull out any PG references.

In the last few cohorts almost everyone has just deployed to Vercel, since it's the simplest path with a Next app. There may have been a couple of Netlify + Supabase/Firebase things too, but never Heroku. Nobody builds Express apps anymore 😅

oliverjam:
Ok should all be updated. We will need to figure out a DB option for the Next.js project, since we previously recommended ElephantSQL, which is Postgres (Vercel doesn't support any native DBs so it has to be a hosted one)

PR #746, opened 2022-10-05
Combine React learning into one resource

### LINKED ISSUE #682: Drop data fetching from React week
Data fetching in React is a super complex topic that is almost impossible to do correctly and ergonomically with simple vanilla React code. The React docs themselves recommend either using a meta-framework like Next.js or a library like react-query for this

Screenshot 2022-06-22 at 12 37 37

Doing this manually in a `useEffect` confuses learners and makes it harder for them to learn about effects, since data fetching is not really their primary use case.

I propose we drop API requests from the workshop/projects and instead have them build stuff with more dynamic client-side interactivity without having to worry about the network. This stuff will then get covered when they learn about Next.js the following week

oliverjam:
Combines "React Components", "React State & Effects" and "React Forms" learning into one resource.

I've edited these down a lot to make the reading more concise; it links to the Beta React Docs in places for more context if people want to dive a bit deeper.

I still need to write the accompanying challenge.

fix #682

### DIFF
--- a/src/course/syllabus/apprenticeship/client-side-app/project.md
+++ b/src/course/syllabus/apprenticeship/client-side-app/project.md
@@ -27,13 +27,14 @@ The goal is to build an interactive game-like thing that uses data from an API.
 
 We'll be using [Vite](https://vitejs.dev/) to handle bundling our app and providing a dev server. You can quickly scaffold a new app using their CLI:
 
-1. `npm init @vitejs/app`
-1. Follow the instructions in your Terminal (pick the `react` option)
-1. `cd [my-app-name]`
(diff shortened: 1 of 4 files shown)


ByteSizedIT (changes requested):
Hey Oli, thanks for this. A couple of notes:

- It seems that each time the Title component function is being defined it returns a div, but the examples calling it show a h1 being returned? 

- There seems to be a typo in the side effects section: 
"However **apps** sometimes your app needs to sync with something else, like fetching from an API or setting up a timer"

oliverjam:
@ByteSizedIT thanks for proof-reading, should both be fixed

ByteSizedIT (approved):

PR #760, opened 2022-10-18
Make Next.js project fit new curriculum.

### LINKED ISSUE #723: Find hosted SQLite DB for Next project
The Next.js projects are deployed to Vercel. This is a serverless platform and so does not support any DBs—data persistence must be hosted somewhere else. For ease we have recommended ElephantSQL previously (really simple free hosted Postgres). Now that we're switching to SQLite it would be a big jump for them to suddenly have to learn PG, especially on top of Next. Ideally we can find a hosted SQLite solution or something (I have no idea how this would even work since SQLite is an in-process DB, but I'll do some research).

### LINKED ISSUE #732: React/Next testing
Now that we've dropped Cypress from the main curriculum we don't have a good testing story for React/Next apps. These are a lot harder to test with the simple Node-based testing we do in the first few weeks, since they involve so much non-standard JS (JSX) or compiler magic (Next routing etc).

Next does have [a page on testing](https://nextjs.org/docs/testing) with various technologies, so in the spirit of that week being about learning to use external resources/teach themselves, we could just link them to it and say "figure out how to test your app".

Alternatively I could try to figure out a really simple/minimal way to test React components without all the heavyweight stuff.

oliverjam:
- [x] Remove references to Postgres.
- [x] Suggest Playwright for testing.
- [x] Suggest Fly for persistent deploys.

My compromise solution to hosting was to give them the option to stick with Heroku if they want to focus on Next and don't mind the lack of DB persistence, otherwise they can try out Fly. The Fly docs are pretty self-explanatory (you run `fly launch` in your app directory and it auto-generates everything you need) so I'm not sure if we need more guidance than that.

Fix #723 
Fix #732

### DIFF
--- a/src/course/syllabus/apprenticeship/full-stack-app/resources.md
+++ b/src/course/syllabus/apprenticeship/full-stack-app/resources.md
@@ -1,2 +1,4 @@
-- [Next.js Docs](https://nextjs.org/docs/getting-started)
+- [How Next.js works](https://nextjs.org/learn/foundations/how-nextjs-works)
 - [Next.js cookie auth example](https://github.com/oliverjam/next-cookie-auth) (including Fly.io deployment)
+- [Deploy Next.js to Fly.io](https://fly.io/docs/languages-and-frameworks/nextjs/)
+- [Persistent Volumes on Fly.io](https://fly.io/docs/reference/volumes/)
--- a/src/course/syllabus/apprenticeship/full-stack-app/resources.md
+++ b/src/course/syllabus/apprenticeship/full-stack-app/resources.md
@@ -1,5 +1,6 @@
 - [How Next.js works](https://nextjs.org/learn/foundations/how-nextjs-works)
 - [Next.js + SQLite example](https://github.com/oliverjam/next-cookie-auth) (including Fly.io deployment)
 - [Deploy Next.js to Fly.io](https://fly.io/docs/languages-and-frameworks/nextjs/)
+- [Testing Next.js](https://nextjs.org/docs/testing)
 - [Persistent Volumes on Fly.io](https://fly.io/docs/reference/volumes/)
-- [Testing Next.js](https://nextjs.org/docs/testing/)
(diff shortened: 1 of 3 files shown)


src/course/syllabus/apprenticeship/full-stack-app/schedule.md
```diff
@@ -2,6 +2,11 @@
 layout: schedule
 schedule:
   monday:
+    - name: Learn Next.js
+      url: https://nextjs.org/learn/basics/create-nextjs-app
+      type: workshop
+      start: 9:45
+      end: 18:00
   tuesday:
```
ByteSizedIT, line 10: I think the empty challenge slot should be additional project time - am aware last cohort seemed to have the whole day for the project (this time there is TfB scheduled for part of the afternoon).

Given the above, maybe we should also reduce down the discussion time in the morning - although I can do this add hoc on the day rather than changing the schedule

src/course/syllabus/apprenticeship/full-stack-app/resources.md
```diff
@@ -1,3 +1,5 @@
-- [Next.js Docs](https://nextjs.org/docs/getting-started)
-- [Next.js Deployment](https://nextjs.org/docs/deployment)
-- [Next.js cookie auth example](https://github.com/oliverjam/next-cookie-auth)
+- [How Next.js works](https://nextjs.org/learn/foundations/how-nextjs-works)
+- [Next.js + SQLite example](https://github.com/oliverjam/next-cookie-auth) (including Fly.io deployment)
+- [Deploy Next.js to Fly.io](https://fly.io/docs/languages-and-frameworks/nextjs/)
+- [Persistent Volumes on Fly.io](https://fly.io/docs/reference/volumes/)
```
ByteSizedIT, line 4: Maybe add https://oliverjam.es/blog/deploying-to-fly/

ByteSizedIT (approved):

PR #869, opened 2023-05-05
update the schedule for the foundation programme

ivanmauricio:


sofer (approved):

sofer:
Much better! (although it's really no more than a rough outline at this stage)

PR #936, opened 2023-11-23
Expert feedback slots

### LINKED ISSUE #935: During FT programme, make apprentice opportunities for FEEDBACK more obvious
-Due to feedback from Ofsted that 'learners do not receive expert guidance they need to improve on weakness', we want to make it even more obvious from our coursebook where/when learners receive expert feedback and support 
- Could we change the colours of the live code reviews block and preface with 'Expert Feedback: Live Code Reviews' - just to make it even more explicit?
- We could do the same for the Feedback sessions on Fridays every 3 weeks when learners are encouraged to act on the feedback from progress logs
-  Can we also change the week 10 'thought of the week' session to PRESENTATION FEEDBACK as that was how we did it this time round, again with the feedback colour

camelPhonso:
Closes #935 

Edited the default schedule so that all 'Live Code Review' slots are now prefaced with 'Expert Feedback' and are highlighted in the same colour as 'Thought of the Week' slots to correlate any sessions where material or feedback is delivered by the FAC team to the learners.

Expert Feedback - Live Code Review slot

Edited all Progress Log and Progress Log Review sessions to be displayed in the same colour as Live Code Review and Thought of the Week

Progress Log slot

Week 10 Thought of the Week session now displayed as 'Presentation Feedback' and again displayed in the same colour as all other content delivery sessions

Presentation Feedback slot

### DIFF
--- a/src/_data/defaultSchedule.yml
+++ b/src/_data/defaultSchedule.yml
@@ -141,6 +141,7 @@ friday:
   - name: Progress Log
     start: 10:00
     end: 10:30
+    type: expert-feedback
   - name: Workshop
     start: 10:30
     end: 18:00
--- a/src/_data/defaultSchedule.yml
+++ b/src/_data/defaultSchedule.yml
@@ -96,8 +96,9 @@ thursday:
     name: Check-in
   - start: 10:00
     end: 11:00
-    name: Live code review
+    name: Expert Feedback - Live code review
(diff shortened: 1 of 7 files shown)


eliazzo (commented):
Looks great! My only thought is that the word 'expert' puts quite a lot of pressure on CF's if they are covering for Ivan

PR #943, opened 2024-01-23
Update node workshop

eliazzo:
I have updated the node workshops. It's pretty similar to before except that I have replaced cowsay with the almost identical figlet. The output is written to a file instead of in the terminal. I think viewing the behaviour in vs code is more intuitive at this stage (vs in the terminal). There are still opportunities to try running code in the terminal.

Feedback welcome

I have created an archive folder which has the old workshop in it. Following a conversation with @sofer  I will be moving any previous versions of pages into here from now on.

### DIFF
--- a/archive/index.md
+++ b/archive/index.md
@@ -1,0 +1,235 @@
+---
+title: Node & npm introduction
+description: Learn how to use Node and npm to run JS on your machine
+tags:
+  - workshop
+  - js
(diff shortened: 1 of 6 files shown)


camelPhonso (approved):
Seems good to me 👍

ShaughnAnderson94 (approved):

PR #953, opened 2024-04-29
remove topic intro

ivanmauricio:


ShaughnAnderson94 (approved):
looks good to me I do want to note that we should probably archive all this stuff before we merge this to main since we do use it as the basis for the tuesday ongoing support structure

ivanmauricio:
@ShaughnAnderson94 noted! i am planning to make a branch which will keep a snapshot of the repo as it was for the last full time cohort we delivered in autumn 2023

PR #961, opened 2024-05-08
Typescript workshop

oliverjam:
Adds a new workshop covering TypeScript from the perspective of JS devs who haven't used TS much. I haven't added it to the schedule since it looks like you're in the process of finalising that in another PR.

Let me know if you think there's anything that isn't covered that the learners will need (or if I've covered anything that's unnecessary for them at this stage).

The first half covers getting a generic Node codebase working with TS (i.e. compiling and setting up dev environment). The second half follows along building a simple todo app that slowly introduces what I think are the most important TS features.

- [x] Type annotations
   - [x] Function parameter types
   - [x] Function return types
   - [x] Type inference
- [x] Literal types
- [x] Type aliases
   - [x] Object types
   - [x] How they differ from interfaces
- [x] Type unions
   - [x] Discriminated unions
- [x] Collection types, e.g. arrays
   - [x] Using generics
- [x] Type narrowing
- [x] When the type system can't help
- [x] 3rd party type definitions
- [x] Type assertions
- [x] Intersection types
- [x] Generic types

### DIFF
--- a/src/styles/partials/syntax-highlighting.css
+++ b/src/styles/partials/syntax-highlighting.css
@@ -75,6 +75,10 @@ pre.language-jsx::after {
   content: "JSX";
 }
 
+pre.language-ts::after {
+  content: "TS";
+}
+
 .token.comment,
 .token.prolog,
 .token.doctype,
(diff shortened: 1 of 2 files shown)


oliverjam:
I've also create the challenge repo here: https://github.com/foundersandcoders/typescript-challenge

It's a bit minimal but honestly I was struggling to come up with ways to practice TS that weren't just adding types to katas. It sounds like they're going to get lots of practice throughout the course either way, so this is just to reinforce the basics they get in the workshop :)

PR #979, opened 2024-07-04
added workshop

ShaughnAnderson94:
added the workshop I have given it a quick spell check but the content and how clear it is I would like a 2nd set of eyes on. Needs to go up by end of day. Also I probably missed at least one thing in my spell check

### DIFF
--- a/src/learn/dotnet-two/index.md
+++ b/src/learn/dotnet-two/index.md
@@ -1,0 +1,297 @@
+---
+title: Dependency injections and interfaces in Dotnet
+description: Learn how to use dependency injections and interfaces in Dotnet
+tags:
+  - workshop
+keywords:
+  - DOTNET
+  - ASP.NET 
+  - CSharp
+challenge: https://github.com/foundersandcoders/DOTNET-Workshop-Two/blob/main/README.md
+---
+
+
+
+## Setting up .Net
+
+Dot net projects can be made in VSCode just like javascript. In order to get started with this you will need to first install the dot net SDK https://dotnet.microsoft.com/en-us/download and the C# Dev kit https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.csdevkit
+
+
+
+
+## Why Dependency Injections
+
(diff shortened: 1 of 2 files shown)


src/learn/dotnet-two/index.md
```diff
@@ -0,0 +1,297 @@
+---
+title: Dependency injections and interfaces in Dotnet
```
ivanmauricio, line 2: this should be `.NET`

src/learn/dotnet-two/index.md
```diff
@@ -0,0 +1,297 @@
...
+description: Learn how to use dependency injections and interfaces in Dotnet
+tags:
+  - workshop
+keywords:
+  - DOTNET
```
ivanmauricio, line 7: i guess this should be `.NET` too

src/learn/dotnet-two/index.md
```diff
@@ -0,0 +1,297 @@
...
+  - ASP.NET 
+  - CSharp
+challenge: https://github.com/foundersandcoders/DOTNET-Workshop-Two/blob/main/README.md
+---
+
+
+
+## Setting up .Net
```
ivanmauricio, line 15: I reckon maybe search the whole repo for dotnet, .net (case insensitive) and change them all to .NET. 

maybe not every single one because you might be referring to CLI package but probably most of them.

src/learn/dotnet-two/index.md
```diff
@@ -0,0 +1,297 @@
...
+
+Dot net projects can be made in VSCode just like javascript. In order to get started with this you will need to first install the dot net SDK https://dotnet.microsoft.com/en-us/download and the C# Dev kit https://marketplace.visualstudio.com/items?itemName=ms-dotnettools.csdevkit
```
ivanmauricio, line 17: another variation `dot net`

src/learn/dotnet-two/index.md
```diff
@@ -0,0 +1,297 @@
...
+```
+We now need to install Microsoft EntityFrameworkCore to help set up a quick in memory database. Like so
+
+
+```dotnet
+
+dotnet add package Microsoft.EntityFrameworkCore
+
+```
+then we make a data folder and inside make a BookContext.cs file that will look like so
```
ivanmauricio, line 54: there's a few capitalisation errors and other small editing things. I think the easiest thing is to get ai to correct it. here it is!

https://ivan-takes-notes.notion.site/NET-PR-edits-1b027250634f4b7b8fe2f71e8145b838?pvs=4

ivanmauricio (changes requested):
thank you for doing this! i've added a few comments to improve clarity and formatting.

Issue #991, opened 2024-09-16
Nested solution in workshop/functions-callbacks-async


bescka:
The current workshop exercise for functions-callbacks-async, [here](https://learn.foundersandcoders.com/workshops/functions-callbacks-async/), uses a highly nested solution to demonstrate the concept of asynchronous callbacks. Given the course prerequisites to have completed Execute: Arrays, it might be informative to add a stretch solution which extends the problem and explores the use of arrays. This could highlight some interesting properties of setTimeout() when used with iterators and also cement the idea of using index variables. 

### Workshop

Let's use callbacks to make some traffic lights. Download the starter files using the command at the top of this workshop. Open `challenge/index.html` in your editor.

1. Inside the `script` tag write a function `light` that takes two parameters: a string and a callback
1. It should wait 1 second, log the string and then call its callback parameter
1. Use `light` to log each colour of a traffic light sequence, in order, followed by `"finished"`
   - e.g. It should log:
   ```
   green
   amber
   red
   amber
   green
   finished
   ```
   with a 1 second pause before each colour

{% box %}

**Hint**: to runs async tasks one after the other remember that you can _nest_ function calls.

{% endbox %}

---

### Solution: 

Current Solution: 
```js
      function light(colour, callback) {
        setTimeout(() => {
          console.log(colour);
          if (callback) {
            callback();
          }
        }, 1000);
      }

      light("green", () => {
        light("amber", () => {
          light("red", () => {
            light("amber", () => {
              light("green", () => {
                console.log("finished");
              });
            });
          });
        });
      });
```

### Stretch

Now we have a nested function that logs our traffic lights. However, highly nested functions can be hard to read. We can mitigate this by using a loop. Open `challenge/stretch/index.html` in your editor.

1. Define an array, `light_sequence` containing each colour of the light sequence, ending with "finished"
2. Modify your `light` function to include a third delay parameter and pass it to your `setTimeout` function
3. Use a method such as `forEach` on `light_sequence` to pass each element of the array to the `light` function
4. Start by calling the `light` function with a `1` second pause. What do you notice?
5. Try and replicate the behaviour of the previous function exactly. How can you ensure the delay is kept between the log of each colour?

{% box %}

**Hint**: _forEach_ can take an additional positional argument.

{% endbox %}

---

### Solution 
```js     
const light_sequence = [
        "green",
        "amber",
        "red",
        "amber",
        "green",
        "finished",
      ];

function light(str, callback, delay) {
[...]

PR #994, opened 2024-09-20
Add stretch to async workshop

bescka:
Issue #991 

Added array method to async workshop task.


AlexVOiceover (approved):
Great addition to the workshop, it's important to show the students different angles to solve a problem

